PR #4, opened 2019-12-09
Binary Ninja support

kumarak:
Supports for binary ninja and uses dwarf debug information to get the type signature


python/anvill/dwarf.py
```diff
@@ -0,0 +1,468 @@
+
```
pgoodman: Copyright on this file?

python/anvill/dwarf.py
```diff
@@ -0,0 +1,468 @@
...
+import enum
+
+from collections import defaultdict, OrderedDict
+from collections import namedtuple
+
+try:
+  from elftools.dwarf.descriptions import (describe_DWARF_expr, set_global_machine_arch, describe_form_class)
+  from elftools.elf.elffile import ELFFile
+except ImportError:
+  print "Install pyelf tools"
```
pgoodman: use `print()` for python3 compat. Perhaps re-`raise`

python/anvill/ida.py
```diff
@@ -38,6 +38,7 @@
 from .os import *
 from .program import *
 from .type import *
+from .dwarf import *
```
pgoodman: lets wrap this in a `try`...`except` block, and maybe recover, providing some kind of fake `DWARFCore` class that is a big NOP or something? I would like this to gracefully degrade, if possible.

python/anvill/ida.py
```diff
@@ -605,6 +606,7 @@ class IDAProgram(Program):
   def __init__(self, *args):
     super(IDAProgram, self).__init__(_get_arch(), _get_os())
     self._functions = weakref.WeakValueDictionary()
+    self._dwarf = DWARFCore(idc.GetInputFilePath())
```
pgoodman: Probably need to `try`...`except` in the event that this isn't an ELF file, support graceful degradation.

scripts/export_json.py
```diff
@@ -0,0 +1,247 @@
+
```
pgoodman: Copyright?

scripts/export_json.py
```diff
@@ -0,0 +1,247 @@
...
+log = logging.getLogger(LOGNAME)
+
+_TYPE_MAP= {
+  'int8_t'  : 'b',
+  'uint8_t' : 'B',
+  'int16_t' : 'h',
+  'uint16_t': 'H',
+  'int32_t' : 'i',
+  'uint32_t': 'I',
+  'long'    : 'l',
```
pgoodman: `int64_t`, `uint64_t`, `float`, `double`, `char`, `short`, etc. ?

scripts/export_json.py
```diff
@@ -0,0 +1,247 @@
...
+  for x in xrange(offset, offset+length):
+    br.seek(x)
+    bytestr.append('{:x}'.format(br.read8()))
+  return ''.join(bytestr)
+
+# Utility wrappers
+def address(addr):
+  return addr
+
+class XRef(object):
```
pgoodman: This class seems to go unused.

scripts/export_json.py
```diff
@@ -0,0 +1,247 @@
...
+    self._json_object['return_values'] = self.get_return_registers()
+
+  def get_params(self):
+    param_list = []
+    index = 0
+    param_regs = self._cc.int_arg_regs
+    for var in self._func.parameter_vars:
+      param = {}
+      param['name'] = str(var)
+      param['type'] = _TYPE_MAP[str(var.type)]
```
pgoodman: What if `str(var.type)` isn't in the type map? This seems like it could be very likely. Perhaps use your code that translates binja types to anvill `Type`s, then use a method on that class to serialize to a string.

scripts/export_json.py
```diff
@@ -0,0 +1,247 @@
...
+      elif var.source_type == binja.VariableSourceType.StackVariableSourceType:
+        stack = {}
+        stack['register'] = self._bv.arch.stack_pointer
+        stack['offset'] = var.storage
+        param['memory'] = stack
+      index = index+1
+      param_list.append(param)
+    return param_list
+
+  def get_return_address(self):
```
pgoodman: You might be able to fall back on the defaults already provided by `arch`, as that will handle x86 storing the return address in memory, and aarch64 storing it in a register.

scripts/export_json.py
```diff
@@ -0,0 +1,247 @@
...
+    regs_list = []
+    for r in self._func.return_regs:
+      reg = {}
+      reg['register'] = r
+      reg['type'] = str(self._func.return_type)
+      regs_list.append(reg)
+    return regs_list
+
+  def get_function_raw(self):
+    return '\x00'
```
pgoodman: ?

scripts/export_json.py
```diff
@@ -0,0 +1,247 @@
...
+  def __init__(self, bv):
+    self._bv = bv
+
+def get_stack_info(bv):
+  stack = {}
+  if bv.arch.name == 'x86_64':
+    stack['address'] = address(0x7ffeefbff000)
+    stack['size'] = 0x8000
+    stack['start_offset'] = 0x1000
+  return stack
```
pgoodman: And what about for other architectures?

scripts/export_json.py
pgoodman: I have made comments on this file, but I think you've actually deprecated this code, so we may as well remove this code if it's not used anymore.

PR #6, opened 2019-12-18
Check llvm::Expected as is now required before destruction

Azeezah:


lib/Lift.cpp
```diff
@@ -251,7 +256,11 @@ class TraceManagerImpl final : public TraceManager {
 bool TraceManagerImpl::TryReadExecutableByte(uint64_t addr, uint8_t *byte) {
   if (auto addr_byte = program.FindByte(addr)) {
     if (addr_byte.IsExecutable()) {
-      *byte = *addr_byte.Value();
+      auto val = addr_byte.Value();
+      if (!val) {
+        LOG(FATAL) << "Found invalid address of executable byte";
```
pgoodman: Here we'll want to `return false`.

pgoodman: I.e. this isn't a fatal error.l

Azeezah: Fixed.

PR #12, opened 2020-01-09
Bitcode Specifier

aaronyoo:


include/anvill/Arch.h
```diff
@@ -0,0 +1,186 @@
...
+  RegisterConstraint(std::vector<VariantConstraint> _variants)
+      : variants(_variants) {}
+
+  std::vector<VariantConstraint> variants;
+};
+
+class CallingConvention {
+ public:
+  CallingConvention(llvm::CallingConv::ID _identity) : identity(_identity) {}
+  virtual ~CallingConvention(){};
```
pgoodman: `= default;` in place of `{}`.

include/anvill/Arch.h
```diff
@@ -0,0 +1,186 @@
...
+class X86_64_SysV : public CallingConvention {
+ public:
+  X86_64_SysV() : CallingConvention(llvm::CallingConv::X86_64_SysV) {}
+  ~X86_64_SysV(){};
+  std::vector<ParameterDecl> BindParameters(const llvm::Function &function);
+  std::vector<ValueDecl> BindReturnValues(const llvm::Function &function);
+  remill::Register *BindReturnStackPointer(const llvm::Function &function);
+
+ private:
+  // TODO: add the rest of the Registers here, not just the 32-bit and 64-bit
```
pgoodman: All `TODO` comments should have a username in them, e.g. `TODO(aaronyoo): blah blah` so then someone looking at the code knows who to ask about the context of that TODO.

include/anvill/Arch.h
```diff
@@ -0,0 +1,186 @@
...
+  // variants
+  const std::vector<RegisterConstraint> parameter_register_constraints = {
```
pgoodman: Could this be a `std::array`, or even just a plain old array?

include/anvill/Arch.h
```diff
@@ -0,0 +1,186 @@
...
+
+      RegisterConstraint({VariantConstraint("XMM0", kTypeFloat, kMaxBit128)}),
+      RegisterConstraint({VariantConstraint("XMM1", kTypeFloat, kMaxBit128)}),
+      RegisterConstraint({VariantConstraint("XMM2", kTypeFloat, kMaxBit128)}),
+      RegisterConstraint({VariantConstraint("XMM3", kTypeFloat, kMaxBit128)}),
+      RegisterConstraint({VariantConstraint("XMM4", kTypeFloat, kMaxBit128)}),
+      RegisterConstraint({VariantConstraint("XMM5", kTypeFloat, kMaxBit128)}),
+      RegisterConstraint({VariantConstraint("XMM6", kTypeFloat, kMaxBit128)}),
+      RegisterConstraint({VariantConstraint("XMM7", kTypeFloat, kMaxBit128)}),
+  };
```
pgoodman: Can `XMM8` et al. be used?

include/anvill/Arch.h
```diff
@@ -0,0 +1,187 @@
...
+  kMinBit32 = kBit128 | kBit64 | kBit32,
+};
+
+enum TypeConstraint : unsigned {
+  kTypeInt = (1 << 0),
+  kTypePtr = (1 << 1),
+  kTypeFloat = (1 << 2),
+  kTypeVec = (1 << 3),
+
+  kTypeIntegral = kTypeInt | kTypePtr,
```
pgoodman: Recommend using `llvm::DataLayout::isNonIntegralPointerType` when trying to pass a pointer, and then if it doesn't fit, have a `LOG(FATAL)` or something, and then have some kind of message related that, and file an issue related to it, and then reference the issue # in the fatal message, asking to basically tell us, via the issue, about code that fails this.

lib/Arch.cpp
```diff
@@ -0,0 +1,441 @@
...
+  for (unsigned int i = 1; i <= num_args; i++) {
+    if (!param_names.count(i)) {
+      param_names[i] = "param" + std::to_string(i);
+    }
+  }
+
+  return param_names;
+}
+
+// TODO: This doesn't really fit here so I will need to find a better place for
```
pgoodman: `TODO(aaronyoo):`

lib/Arch.cpp
```diff
@@ -0,0 +1,441 @@
...
+    }
+    case llvm::Type::X86_MMXTyID: {
+      ret = "M";
+      break;
+    }
+    case llvm::Type::IntegerTyID: {
+      auto derived = llvm::cast<llvm::IntegerType>(type);
+      // Since there is no way to check for uint vs. int, lower all integer
+      // types to signed integers. Maybe come back when there is a better way to
+      // figure out unsigned vs signed ints
```
pgoodman: Could look for usages, e.g. is it used in an `add`, and if so, is it signed or unsigned overflow attributed, or `div` vs. `sdiv` kind of thing. File an issue for it, add a TODO comment referencing the issue.

lib/Arch.cpp
```diff
@@ -0,0 +1,441 @@
...
+        }
+        case 64: {
+          ret = sign ? "l" : "L";
+          break;
+        }
+      }
+      break;
+    }
+    case llvm::Type::FunctionTyID: {
+      ret = "func";
```
pgoodman: This isn't right. Function types are parentheses, e.g. `(ii)` is a function taking an `int` and returning an `int`. `(vi)` is taking void, returning int. `(&v)` is a varargs func returning void. You'll need recursion here.

lib/Arch.cpp
```diff
@@ -0,0 +1,441 @@
...
+      break;
+    }
+    case llvm::Type::StructTyID: {
+      auto struct_ptr = llvm::cast<llvm::StructType>(&type);
+      std::string element_list = "";
+      for (unsigned i = 0; i < struct_ptr->getNumElements(); i++) {
+        // TODO: BUG: this if statement protects against infinite chained
```
pgoodman: This needs to be solved. The way to solve it is to pass a "cache" of types down as an argument to this function. We know that structs are the only such type than can enter into this situation. You can see this pattern used here, where `ids` is the cache: https://github.com/lifting-bits/anvill/blob/master/python/anvill/type.py#L240-L253

lib/Arch.cpp
pgoodman: Consult this: https://github.com/lifting-bits/anvill/blob/master/python/anvill/type.py#L444-L460

lib/Arch.cpp
```diff
@@ -0,0 +1,441 @@
...
+    case llvm::Type::PointerTyID: {
+      ret = "*";
+      auto derived = llvm::dyn_cast<llvm::PointerType>(&type);
+      // Get the type of the pointee
+      ret += TranslateType(*derived->getElementType());
+      break;
+    }
+
+    default:
+      LOG(ERROR) << "Could not translate TypeID: " << id;
```
pgoodman: We can do a better job here. I recommend passing down a data layout, then you can ask the data layout for the type storage size, and then fill in an array of bytes type here.

lib/Arch.cpp
```diff
@@ -0,0 +1,441 @@
...
+    }
+  }
+
+  return param_names;
+}
+
+// TODO: This doesn't really fit here so I will need to find a better place for
+// this. Translates an llvm::Type to a type that conforms to the spec in
+// TypeParser.cpp
+std::string TranslateType(const llvm::Type &type) {
```
pgoodman: Instead of returning a `std::string`, I recommend having this function take in a `std::stringstream` by reference and inserting into that. Also recommend taking in a data layout and a unordered map of struct type pointers to integer IDs.

If you want an interface that returns a std::string, then have that as a second function, wrapping this one, where it takes a module pointer and a type pointer, returns a struct, and calls this function.

pgoodman: Maybe make a `TypePrinter.cpp` or `TypeUnparser.cpp` or `TypeSerializer.cpp`

lib/Arch.cpp
```diff
@@ -0,0 +1,441 @@
...
+    llvm::Type &type, std::vector<bool> &reserved,
+    const std::vector<RegisterConstraint> &register_constraints) {
+  SizeConstraint size_constraint;
+  TypeConstraint type_constraint;
+
+  if (type.isIntegerTy()) {
+    type_constraint = kTypeInt;
+    auto derived = llvm::cast<llvm::IntegerType>(type);
+    unsigned int width = derived.getBitWidth();
+    if (width == 64) {
```
pgoodman: This needs to be made more general. I recommend using `llvm::DataLayout::fitsInLegalInteger` to figure out if we can likely fit it somewhere. Then instead of asking for the bit width of the type, ask for the type store size in bits or bytes, and match that to either 32 or 64.

Also, investigate what happens if you have a 128-bit integer passed as an argument or returned.

lib/Arch.cpp
```diff
@@ -0,0 +1,441 @@
...
+    size_constraint = kMinBit32;
+  } else if (type.isDoubleTy()) {
+    type_constraint = kTypeFloat;
+    // We automatically know it is 64-bit IEEE floating point type
+    size_constraint = kMinBit64;
+  } else if (type.isPointerTy()) {
+    type_constraint = kTypeIntegral;
+    size_constraint = kMinBit64;
+  }
+  // TODO: Handle other types
```
pgoodman: `TODO(aaronyoo):`.

You should handle `long double`s, i.e. the `X86_FP80` type here. Possibly also the MMX type.

pgoodman: If you leave something undone, add a `LOG(ERROR)` or in this case, maybe even a `LOG(FATAL)`. Right now this is silently ignoring, which is not good because we won't know when we're missing stuff.

lib/Arch.cpp
```diff
@@ -0,0 +1,441 @@
...
+// Try to allocate a register for the argument based on the register constraints
+// and what has already been reserved. Return nullptr if there is no possible
+// register allocation.
+remill::Register *TryRegisterAllocate(
+    llvm::Type &type, std::vector<bool> &reserved,
+    const std::vector<RegisterConstraint> &register_constraints) {
+  SizeConstraint size_constraint;
+  TypeConstraint type_constraint;
+
+  if (type.isIntegerTy()) {
```
pgoodman: Maybe do a `switch` on the type ID, that'll make some things more explicit, and then you can better group the unhandled cases in terms of fatal unhandled, or non-fatal unhandled.

lib/Arch.cpp
```diff
@@ -0,0 +1,441 @@
...
+  std::vector<bool> allocated(parameter_register_constraints.size(), false);
+
+  // Stack position of the first argument
+  unsigned int stack_offset = 16;
+
+  for (auto &argument : function.args()) {
+    anvill::ParameterDecl declaration = {};
+    declaration.type = argument.getType();
+
+    // Try to allocate from a register. If a register is not available then
```
pgoodman: This doesn't seem to handle the case where a small struct argument passed by value in the bitcode is split across several registers. In that event, we should handle it. When splitting arguments, perhaps append on suffixes, e.g. `_0`, `_1`, etc. to the param names. Make sure not to go out-of-bounds.

lib/Arch.cpp
```diff
@@ -0,0 +1,441 @@
...
+std::vector<anvill::ParameterDecl> X86_64_SysV::BindParameters(
+    const llvm::Function &function) {
+  std::vector<anvill::ParameterDecl> parameter_declarations;
+  auto param_names = TryRecoverParamNames(function);
+  llvm::DataLayout dl(function.getParent());
+
+  // Used to keep track of which registers have been allocated
+  std::vector<bool> allocated(parameter_register_constraints.size(), false);
+
+  // Stack position of the first argument
```
pgoodman: Larger comment describing stack spilling behavior of arguments. Perhaps some small ASCII diagrams or just tabular layout so that it's documented in what order arguments passed on the stack are pushed on the stack.

lib/Arch.cpp
```diff
@@ -0,0 +1,441 @@
...
+    declaration.type = argument.getType();
+
+    // Try to allocate from a register. If a register is not available then
+    // allocate from the stack.
+    if (remill::Register *reg = TryRegisterAllocate(
+            *argument.getType(), allocated, parameter_register_constraints)) {
+      declaration.reg = reg;
+    } else {
+      remill::Register *mem_reg =
+          new remill::Register("RSP", stack_offset, 8, 0, argument.getType());
```
pgoodman: This doesn't make sense. You should never allocate a new `remill::Register`, as it will leak; you should look up the stack pointer from the `Arch`. You're passing a `stack_offset` to fill in the `offset` [1] field, which makes no sense, as that stack offset is unrelated to the `State` structure. You want a `ValueDecl` that has `mem_reg` as the stack pointer, and `mem_offset` as your `stack_offset`. See here: https://github.com/lifting-bits/anvill/blob/master/include/anvill/Decl.h#L61-L62

[1] https://github.com/lifting-bits/remill/blob/master/remill/Arch/Arch.h#L53

lib/Arch.cpp
```diff
@@ -0,0 +1,441 @@
...
+  // For the X86_64_SysV ABI, it is always:
+  //
+  // "return_stack_pointer": {
+  //     "offset": "8",
+  //     "register": "RSP",
+  //     "type": "L"
+  // }
+
+  auto int64_ptr_ty = llvm::IntegerType::get(function.getContext(), 64);
+  return new remill::Register("RSP", 8, 8, 0, int64_ptr_ty);
```
pgoodman: Per comment above, never allocate a new `remill::Register`; you want to look up the stack pointer in the `Arch`, and then use `mem_reg` and `mem_offset`.

lib/Arch.cpp
```diff
@@ -0,0 +1,441 @@
...
+  // Stack position of the first argument
+  unsigned int stack_offset = 4;
+
+  for (auto &argument : function.args()) {
+    anvill::ParameterDecl declaration = {};
+    declaration.type = argument.getType();
+
+    // Since there are no registers, just allocate from the stack
+    remill::Register *mem_reg =
+        new remill::Register("ESP", stack_offset, 4, 0, argument.getType());
```
pgoodman: See above comment about allocating `remill::Register` and using the memory form of `ValueDecl`.

lib/Arch.cpp
```diff
@@ -0,0 +1,441 @@
...
+  return new remill::Register("RSP", 8, 8, 0, int64_ptr_ty);
+}
+
+std::vector<ParameterDecl> X86_C::BindParameters(
+    const llvm::Function &function) {
+  std::vector<anvill::ParameterDecl> parameter_declarations;
+  auto param_names = TryRecoverParamNames(function);
+  llvm::DataLayout dl(function.getParent());
+
+  // Stack position of the first argument
```
pgoodman: More elaborate comment on arg passing behavior on the stack, including order of arguments.

lib/Arch.cpp
```diff
@@ -0,0 +1,441 @@
...
+  unsigned int stack_offset = 4;
+
```
pgoodman: Make sure to check for explicit structure return: https://code.woboq.org/llvm/llvm/include/llvm/IR/Function.h.html#576

If it is there, then the function's return pointer value needs to be adjusted: https://gcc.godbolt.org/z/wZSHuw

Also, check for implicit struct ret, where the return value is a non-void struct, and so you need to pass in the parameter.

lib/Arch.cpp
```diff
@@ -0,0 +1,441 @@
...
+  std::vector<anvill::ParameterDecl> parameter_declarations;
+  auto param_names = TryRecoverParamNames(function);
+  llvm::DataLayout dl(function.getParent());
+
+  // Used to keep track of which registers have been allocated
+  std::vector<bool> allocated(parameter_register_constraints.size(), false);
+
+  // Stack position of the first argument
+  unsigned int stack_offset = 16;
+
```
pgoodman: You need to check for implicit struct ret here, where we have non-clang sourced bitcode returning a big struct that needs to be handled as a parameter. This tells me a minor amount of re-architecting is needed that allows you to try to allocate the return values first, and if not, tell the parameter binding about that, and also tell return stack pointer binding about that (in the 32-bit case).

lib/Arch.cpp
```diff
@@ -0,0 +1,441 @@
...
+    // Get a name for the IR parameter.
+    // Need to add 1 because param_names uses logical numbering, but
+    // argument.getArgNo() uses index numbering.
+    declaration.name = param_names[argument.getArgNo() + 1];
+    parameter_declarations.push_back(declaration);
+  }
+
+  return parameter_declarations;
+}
+
```
pgoodman: Many comments above apply to below, so stopping my review here for now.

BITCODE.cpp
```diff
@@ -0,0 +1,124 @@
...
+    exit(EXIT_FAILURE);
+  }
+
+  std::string arch = triple[0];
+  std::string os = triple[2];
+
+  // Change the architecture into something that the rest of anvill and remill
+  // can understand.
+  if (arch.find("x86_64") != std::string::npos) {
+    arch = "amd64";
```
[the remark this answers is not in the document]
aaronyoo: Fix 837bee6

BITCODE.cpp
```diff
@@ -0,0 +1,124 @@
...
+  }
+
+  // Trim the versioning information
+  // TODO: cover the other OSes as well
+  if (os.find("macos") != std::string::npos) {
+    os = "macos";
+  }
+
+  return std::make_tuple(arch, os);
```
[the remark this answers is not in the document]
aaronyoo: Fix 837bee6

BITCODE.cpp
```diff
@@ -0,0 +1,124 @@
...
+  json.insert(llvm::json::Object::KV{llvm::json::ObjectKey("arch"), arch});
+  json.insert(llvm::json::Object::KV{llvm::json::ObjectKey("os"), os});
+
+  llvm::json::Array funcs_json;
+
+  for (auto &function : *module) {
+    // Skip llvm dbg functions for now
+    // TODO: find a way to deal with this
+    std::string function_name = llvm::demangle(function.getName().data());
+    if (function_name.find("llvm.") == 0) continue;
```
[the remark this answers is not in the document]
aaronyoo: Fix 
3e475d0

include/anvill/Arch.h
aaronyoo: No, don't think so: https://godbolt.org/z/7w-NLA

lib/Arch.cpp
```diff
@@ -0,0 +1,365 @@
...
+  return ret;
+}
+
+std::vector<anvill::ValueDecl> X86_64_SysV::BindReturnValues(
+    const llvm::Function &function) {
+  std::vector<anvill::ValueDecl> return_value_declarations;
+
+  for (auto &block : function) {
+    for (auto &inst : block) {
+      // TODO(aty): There has to be a better way of doing this, right now I am
```
pgoodman: I think this would all be simpler if you used `function.getReturnType()`.

include/anvill/Arch.h
aaronyoo: IMO it is cleaner to use a vector (or at least that is what I have used in the past), and performance is comparable as long as we can avoid copying, which we do. Using an array and passing a pointer to it is the alternative that I was considering. Let me know what you think.

lib/Arch.cpp
aaronyoo: I made changes in 0bbd0d9 that should help the issue. Basically I have two interfaces, one that takes a Datalayout and one that doesn't. I didn't really make sense to me to just have one that accepts a Datalayout because then it would be difficult to serialize a ValueDecl in isolation. So if there is a Datalayout, we translate an unknown type into an array of reasonable size, otherwise, we fail out.

pgoodman: Thinking over more, I think `DataLayout` should be a requirement, and we can "force" it using `const llvm::DataLayout &` as a parameter type. I like reference parameters when you want to communicate "never null" for a pointer.

include/anvill/Arch.h
pgoodman: OK!

lib/Arch.cpp
aaronyoo: Fixed 327d728

lib/Arch.cpp
aaronyoo: Fixed 327d728

lib/Arch.cpp
aaronyoo: Left a TODO for myself to handle MMX types 327d728

lib/Arch.cpp
aaronyoo: Fixed 327d728

lib/Arch.cpp
aaronyoo: Fixed 63d92a6

lib/Arch.cpp
aaronyoo: Clone of above

lib/Arch.cpp
aaronyoo: Clone of above, resolving

BITCODE.cpp
```diff
@@ -0,0 +1,79 @@
...
+
+  // Insert functions array into top level JSON
+  json.insert(llvm::json::Object::KV{llvm::json::ObjectKey("functions"),
+                                     llvm::json::Value(std::move(funcs_json))});
+
+  // Print JSON
+  llvm::raw_fd_ostream S(STDOUT_FILENO, false);
+  S << llvm::formatv("{0:4}", llvm::json::Value(std::move(json)));
+
+  return 0;
```
pgoodman: `EXIT_SUCCESS`.

BITCODE.cpp
```diff
@@ -0,0 +1,79 @@
...
+  // Overwrite the inherited architecture and os flags if they are not
+  // already empty.
+  if (!FLAGS_arch.empty() || !FLAGS_os.empty()) {
+    FLAGS_arch = "";
+    FLAGS_os = "";
+  }
+
+  auto context = new llvm::LLVMContext;
+  auto module = remill::LoadModuleFromFile(context, FLAGS_bc_file);
+  auto arch = remill::Arch::GetModuleArch(*module);
```
pgoodman: Can this fail and return `nullptr`? If so, maybe check for it, log a message, and `return EXIT_FAILURE;`

BITCODE.cpp
```diff
@@ -0,0 +1,79 @@
...
+  llvm::json::Object json;
+  json.insert(llvm::json::Object::KV{llvm::json::ObjectKey("arch"), arch_name});
+  json.insert(llvm::json::Object::KV{llvm::json::ObjectKey("os"), os_name});
+
+  llvm::json::Array funcs_json;
+
+  for (auto &function : *module) {
+    // Skip llvm debug intrinsics
+    const llvm::StringRef func_name = function.getName();
+    if (func_name.startswith("llvm.")) {
```
pgoodman: Another way of doing this is:
```c++
if (function.getIntrinsicId()) {
  continue;
}
```

aaronyoo: Fixed c9cdb7b

BITCODE.cpp
aaronyoo: Fixed c9cdb7b

BITCODE.cpp
aaronyoo: Fixed c9cdb7b

aaronyoo:
I'm still working on the explicit and implicit sret for x86_64_SysV but I wanted to push this so that you could review it.

include/anvill/Arch.h
```diff
@@ -0,0 +1,233 @@
...
+
+enum SizeConstraint : unsigned {
+  kBit8 = (1 << 0),
+  kBit16 = (1 << 1),
+  kBit32 = (1 << 2),
+  kBit64 = (1 << 3),
+  kBit80 = (1 << 4),
+  kBit128 = (1 << 5),
+
+  kMaxBit128 = kBit128 | kBit80 | kBit64 | kBit32 | kBit16 | kBit8,
```
pgoodman: I'm not sure about `kBit80` being in here. I don't believe we can pass `long double`s over `xmm` regs.

include/anvill/Arch.h
```diff
@@ -0,0 +1,233 @@
...
+
+  std::string register_name;
+  TypeConstraint type_constraint;
+  SizeConstraint size_constraint;
+};
+
+// Captures the different sizes and constraints of a single register
+struct RegisterConstraint {
+  RegisterConstraint(std::vector<VariantConstraint> _variants)
+      : variants(_variants) {}
```
pgoodman: `std::move(_variants)`?

include/anvill/Arch.h
```diff
@@ -0,0 +1,233 @@
...
+      const llvm::StructType &st,
+      const std::vector<RegisterConstraint> &constraints);
+
+  const remill::Arch* arch;
+
+ private:
+  llvm::CallingConv::ID identity;
+};
+
+class X86_64_SysV : public CallingConvention {
```
pgoodman: I think this can go into a .cpp file.

include/anvill/Arch.h
```diff
@@ -0,0 +1,233 @@
...
+      RegisterConstraint({VariantConstraint("XMM1", kTypeFloat, kMaxBit128)}),
+
+      // Since the FPU registers are 80 bits wide, they are only able to hold
+      // 64-bit values.
+      RegisterConstraint({VariantConstraint("ST0", kTypeFloat, kMaxBit80)}),
+      RegisterConstraint({VariantConstraint("ST1", kTypeFloat, kMaxBit80)}),
+  };
+};
+
+// This is the cdecl calling convention referenced by llvm::CallingConv::C
```
pgoodman: I think this can go into a .cpp file.

include/anvill/Arch.h
```diff
@@ -0,0 +1,233 @@
...
+          VariantConstraint("R8", kTypeIntegral, kMaxBit64),
+      }),
+      RegisterConstraint({
+          VariantConstraint("R9L", kTypeIntegral, kMaxBit8),
+          VariantConstraint("R9W", kTypeIntegral, kMaxBit16),
+          VariantConstraint("R9D", kTypeIntegral, kMaxBit32),
+          VariantConstraint("R9", kTypeIntegral, kMaxBit64),
+      }),
+
+      RegisterConstraint({VariantConstraint("XMM0", kTypeFloat, kMaxBit128)}),
```
pgoodman: Vectors of floats/integers can be passed over xmm regs in 64 bit. https://gcc.godbolt.org/z/BfgENx

include/anvill/Arch.h
```diff
@@ -0,0 +1,233 @@
...
+          VariantConstraint("EDX", kTypeIntegral, kMaxBit32),
+      }),
+      RegisterConstraint({
+          VariantConstraint("CL", kTypeIntegral, kMaxBit8),
+          VariantConstraint("CX", kTypeIntegral, kMaxBit16),
+          VariantConstraint("ECX", kTypeIntegral, kMaxBit32),
+      }),
+      RegisterConstraint({VariantConstraint("ST0", kTypeFloat, kMaxBit80)}),
+      RegisterConstraint({VariantConstraint("ST1", kTypeFloat, kMaxBit80)}),
+  };
```
pgoodman: Here's some confusing things, probably deserving of some comments in the code:
Vectors (specified in C) of integers/floats can be passed over mmx registers in 32-bit with the gcc, e.g. https://gcc.godbolt.org/z/X9JjUV (gcc), though Clang will use memory (https://gcc.godbolt.org/z/8QX3eW) instead. This seems to be because Clang do ABI lowering of an `<i32, i32>` or `<float, float>` to either an `i64` or a `double`, depending on 32-bit or 64-bit. If we try to force Clang's hand, by manually using vector valued parameters, then it uses XMM registers: https://gcc.godbolt.org/z/C9Ppfg and https://gcc.godbolt.org/z/PH5Q__ for parameter passing.

pgoodman: We'll probably want to handle the latter case, of being able to put floats/integer vector values into xmm regs.

pgoodman: Maybe someone will respond: https://twitter.com/peter_a_goodman/status/1217942699594207232

lib/Decl.cpp
```diff
@@ -188,4 +195,135 @@ llvm::Value *FunctionDecl::CallFromLiftedBlock(
...
+
+  // Try to guess the parameter and return value register allocation by looking
+  // at the calling convention of the function
+  llvm::CallingConv::ID cc_id = func.getCallingConv();
+  std::unique_ptr<CallingConvention> cc = nullptr;
+
+  switch (cc_id) {
+    case llvm::CallingConv::X86_64_SysV:
+      cc = std::unique_ptr<anvill::X86_64_SysV>(new X86_64_SysV(arch));
+    default:
```
pgoodman: I think you want a `break;` here before the `default:`.

lib/Decl.cpp
```diff
@@ -188,4 +195,135 @@ llvm::Value *FunctionDecl::CallFromLiftedBlock(
...
+      // TODO(aty): find a better way to do this since the calling conventions given
+      // by llvm::function cannot be trusted
+      cc = std::unique_ptr<anvill::X86_C>(new X86_C(arch));
```
pgoodman: I would check the arch name / OS to decide on the most suitable fallback ABI here. It doesn't make sense to use a 32-bit ABI if we're on a 64-bit arch, for example.

include/anvill/Arch.h
aaronyoo: I don't think this allows you to because we still have the type constraint. Since long double would be an integral type it wouldn't be able to be passed over xmm.

lib/Decl.cpp
aaronyoo: Fixed 75dfa9a

include/anvill/Arch.h
aaronyoo: Fixed 75dfa9a

include/anvill/Arch.h
aaronyoo: Fixed 75dfa9a

include/anvill/Arch.h
aaronyoo: Fixed 75dfa9a

include/anvill/Arch.h
aaronyoo: Fixed 75dfa9a

include/anvill/Arch.h
aaronyoo: So, if we decided on the doing the latter case emulating clang, then this was fixed in 
75dfa9a by the same change that lets vectors be passed in xmm regs (see your comment above). I can put a comment if you think its necessary, but if we are emulating clang it seems logical. If clang lowers it for us then we will not try to allocate in xmm

lib/Arch/X86_64_SysV.cpp
```diff
@@ -0,0 +1,158 @@
...
+
+    // Try to allocate from a register. If a register is not available then
+    // allocate from the stack.
+    if (const remill::Register *reg = TryRegisterAllocate(
+            *argument.getType(), allocated, parameter_register_constraints)) {
+      declaration.reg = reg;
+    } else {
+      declaration.mem_offset = stack_offset;
+      declaration.mem_reg = arch->RegisterByName("RSP");
+      stack_offset += dl.getTypeAllocSize(argument.getType());
```
pgoodman: I think we should see if alignment changes anything. For example, can we arrange for a stack-allocated parameter of type `struct alignas(N) { char x; };` with `N` like `1`, `3`, `7`, `11`, `15`, etc. to see if more space is allocated than just what LLVM uses to represent the type.

lib/Arch/X86_C.cpp
```diff
@@ -0,0 +1,152 @@
...
+  // However, even if there is sret on the second argument as well, we do not
+  // need to worry about this. For some reason the callee is only responsible
+  // for cleaning up the case where an sret argument is passed in as the first
+  // argument.
+  if (func.hasParamAttribute(0, llvm::Attribute::StructRet)) {
+    fdecl.return_stack_pointer_offset = 8;
+  } else {
+    fdecl.return_stack_pointer_offset = 4;
+  }
+
```
pgoodman: Below your might do `injected_sret`, though that doesn't seem to influence this code. Perhaps an additional `bool injected_sret` parameter should be passed in. Alternatively, you could have an integer value that you pass in, where the parameter and return value binding code is responsible for figuring out how much callee-cleanup needs to happen.

lib/Arch/X86_C.cpp
```diff
@@ -0,0 +1,152 @@
...
+  value_declaration.type = ret_type;
+
+  switch (ret_type->getTypeID()) {
+    case llvm::Type::IntegerTyID:
+    case llvm::Type::PointerTyID: {
+      // Allocate EAX for an integer or pointer
+      value_declaration.reg = arch->RegisterByName("EAX");
+      break;
+    }
+    case llvm::Type::FloatTyID: {
```
pgoodman: `double`, `log double` can be handled here too.

pgoodman: mmx types can also be handled by a simple case of looking for the `MM0` register.

lib/Arch/X86_C.cpp
```diff
@@ -0,0 +1,152 @@
...
+        // Struct splitting didn't work so do RVO. Assume that the pointer
+        // to the return value resides in EAX. In this case we have injected an
+        // sret into the first parameter so we need to take note of that.
+        injected_sret = true;
+        value_declaration.reg = arch->RegisterByName("EAX");
+      }
+      break;
+    }
+    default: {
+      LOG(ERROR) << "Encountered an unknown return type";
```
pgoodman: Elsewhere you have `LOG(FATAL)`, perhaps use that here. Also, I recommend adding in `remill::LLVMThingToString(ret_type)` to make the logs more actionable.

lib/Decl.cpp
```diff
@@ -188,4 +195,136 @@ llvm::Value *FunctionDecl::CallFromLiftedBlock(
...
+  llvm::json::Object json;
+
+  std::string output_name;
+  if (FLAGS_demangle_names) {
+    output_name = this->demangled_name;
+  } else {
+    output_name = this->name;
+  }
+  json.insert(
+      llvm::json::Object::KV{llvm::json::ObjectKey("name"), output_name});
```
pgoodman: This should always be `this->name`, and so if `FLAGS_demangle_names` is true, and if `!this->demangled_name.empty()`, then add in a new key/value pair to the JSON object, `"demangled_name"`.

lib/Arch/X86_64_SysV.cpp
aaronyoo: I am pretty sure I am using getTypeAllocSize like we talked about, so I will leave this for now.

include/anvill/TypePrinter.h
```diff
@@ -0,0 +1,10 @@
+#pragma once
+
+#include<llvm/IR/DataLayout.h>
```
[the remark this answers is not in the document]
aaronyoo: Fixed 1742dc9

lib/Arch/Arch.cpp
```diff
@@ -0,0 +1,165 @@
...
+
+  switch (type.getTypeID()) {
+    case llvm::Type::IntegerTyID: {
+      type_constraint = kTypeInt;
+      auto derived = llvm::cast<llvm::IntegerType>(type);
+      unsigned int width = derived.getBitWidth();
+      if (width == 64) {
+        size_constraint = kMinBit64;
+      } else {
+        // TODO(aty): I know that this is wrong but for now its fine
```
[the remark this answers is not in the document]
aaronyoo: I now handle i28 and error out fatally if there is something that is a size that we are not ready for. Fix in 1742dc9

lib/Arch/Arch.cpp
```diff
@@ -0,0 +1,165 @@
...
+    default: {
+      LOG(FATAL) << "Could not assign type and size constraints in "
+                    "TryRegisterAllocate()";
+      // TODO(aty): Handle other types like X86_MMXTyID, etc.
+      break;
+    }
+  }
+
+  for (size_t i = 0; i < register_constraints.size(); i++) {
+    if (reserved[i]) continue;
```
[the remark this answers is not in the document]
aaronyoo: Fix 
1742dc9

include/anvill/Arch.h
```diff
@@ -0,0 +1,245 @@
...
+  RegisterConstraint(std::vector<VariantConstraint> _variants)
+      : variants(std::move(_variants)) {}
+
+  std::vector<VariantConstraint> variants;
+};
+
+std::map<unsigned, std::string> TryRecoverParamNames(
+    const llvm::Function &function);
+
+class CallingConvention {
```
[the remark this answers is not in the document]
aaronyoo: Moved the entire Arch.h file into lib since Decl.cpp is the only file using it right now and it should not be exposed as an API. Fix 1742dc9

lib/Arch/Arch.cpp
```diff
@@ -0,0 +1,165 @@
...
+
+    const RegisterConstraint &constraint = register_constraints[i];
+    // Iterate through the different sizes of a single register to find the
+    // smallest possible match
+    for (auto const &variant : constraint.variants) {
+      if (size_constraint & variant.size_constraint &&
+          type_constraint & variant.type_constraint) {
+        reserved[i] = true;
```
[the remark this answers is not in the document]
aaronyoo: Fix 1742dc9. According to the google flags docs, CHECK_NOTNULL does not allow messages like that. I would have to do `CHECKEQ(reg, static_cast<remill::Register*>(nullptr)) << "Message"` So I just decided to go with `if(!reg) ... `

lib/Arch/X86_64_SysV.cpp
```diff
@@ -0,0 +1,158 @@
...
+  value_declaration.type = ret_type;
+
+  switch (ret_type->getTypeID()) {
+    case llvm::Type::IntegerTyID:
+    case llvm::Type::PointerTyID: {
+      // Allocate RAX for an integer or pointer
+      value_declaration.reg = arch->RegisterByName("RAX");
+      break;
+    }
+    case llvm::Type::FloatTyID: {
```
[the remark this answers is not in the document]
aaronyoo: Fix 1742dc9

lib/Arch/X86_64_SysV.cpp
```diff
@@ -0,0 +1,158 @@
...
+        return *mapping;
+      } else {
+        // Struct splitting didn't work so do RVO. Assume that the pointer
+        // to the return value resides in RAX.
+        injected_sret = true;
+        value_declaration.reg = arch->RegisterByName("RAX");
+      }
+      break;
+    }
+    default: {
```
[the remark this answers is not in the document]
aaronyoo: Fix 1742dc9

lib/Arch/X86_C.cpp
aaronyoo: Fix 1742dc9. Passed in injected_sret in this case.

lib/Decl.cpp
aaronyoo: Fixed 028b2a5

lib/Arch/X86_C.cpp
aaronyoo: fixed 028b2a5

lib/Arch/X86_C.cpp
aaronyoo: Fixed 028b2a5

lib/Arch/Arch.cpp
```diff
@@ -0,0 +1,165 @@
...
+namespace anvill {
+
+// Try to recover parameter names using debug information. Otherwise, name the
+// parameters with the form "param_x". The mapping of the return value is
+// positional starting at 1.
+std::map<unsigned, std::string> TryRecoverParamNames(
+    const llvm::Function &function) {
+  std::map<unsigned int, std::string> param_names;
+
+  for (auto &block : function) {
```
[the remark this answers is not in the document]
aaronyoo: Unfortunately, using something like `for(user : arg.users)` doesn't iterate over debug intrinsics. So I guess they are not counted as uses. I will leave this here for now.

lib/Arch/X86_64_SysV.cpp
```diff
@@ -0,0 +1,158 @@
...
+    const llvm::Function &function, bool &injected_sret) {
+  std::vector<anvill::ValueDecl> return_value_declarations;
+  anvill::ValueDecl value_declaration = {};
+  injected_sret = false;
+
+  // If there is an sret parameter then it is a special case. For the
+  // X86_64_SysV ABI, the sret parameters are guarenteed the be in %rax. In this
+  // case, we can assume the actual return value of the function will be the
+  // sret struct pointer.
+  for (auto arg = function.arg_begin(); arg != function.arg_end(); arg++) {
```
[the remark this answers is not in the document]
aaronyoo: Fixed to be made more explicit 028b2a5

lib/Arch/Arch.cpp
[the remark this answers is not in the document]
aaronyoo: Left comment about this behavior

lib/Arch/Arch.cpp
```diff
@@ -0,0 +1,185 @@
...
+    for (auto &inst : block) {
+      if (auto debug_inst = llvm::dyn_cast<llvm::DbgInfoIntrinsic>(&inst)) {
+        if (auto value_intrin = llvm::dyn_cast<llvm::DbgDeclareInst>(&inst)) {
+          const llvm::MDNode *mdn = value_intrin->getVariable();
+          const llvm::DILocalVariable *div =
+              llvm::cast<llvm::DILocalVariable>(mdn);
+
+          // Make sure it is actually an argument
+          if (div->getArg() != 0) {
+            LOG(INFO) << div->getArg() << " : " << div->getName().data();
```
pgoodman: Remove eventually.

lib/Arch/Arch.cpp
```diff
@@ -0,0 +1,185 @@
...
+      break;
+    }
+    case llvm::Type::VectorTyID: {
+      type_constraint = kTypeFloatOrVec;
+      size_constraint = kMinBit80;
+      break;
+    }
+    default: {
+      LOG(FATAL) << "Could not assign type and size constraints in "
+                    "TryRegisterAllocate()";
```
pgoodman: Add in a `remill::LLVMThingToString` into the log message so that it prints out what type wasn't handled.

lib/Arch/Arch.cpp
```diff
@@ -0,0 +1,185 @@
...
+std::unique_ptr<std::vector<anvill::ValueDecl>>
+CallingConvention::TryReturnThroughRegisters(
+    const llvm::StructType &st,
+    const std::vector<RegisterConstraint> &constraints) {
+  auto ret = std::make_unique<std::vector<anvill::ValueDecl>>();
+  std::vector<bool> reserved(constraints.size(), false);
+  for (unsigned i = 0; i < st.getNumElements(); i++) {
+    anvill::ValueDecl value_decl = {};
+    auto reg =
+        TryRegisterAllocate(*st.getElementType(i), reserved, constraints);
```
pgoodman: What if it is a struct in a struct?

pgoodman: Or a struct in a vector, or what not. I think what might help is a kind of visitor, e.g. could be a function taking a type and a callback, and recursively descending down the type hierarchy until a "basic" type is hit.

lib/Arch/Arch.h
```diff
@@ -0,0 +1,270 @@
...
+  kBit80 = (1 << 4),
+  kBit128 = (1 << 5),
+
+  kMaxBit128 = kBit128 | kBit80 | kBit64 | kBit32 | kBit16 | kBit8,
+  kMaxBit80 = kBit80 | kBit64 | kBit32 | kBit16 | kBit8,
+  kMaxBit64 = kBit64 | kBit32 | kBit16 | kBit8,
+  kMaxBit32 = kBit32 | kBit16 | kBit8,
+  kMaxBit16 = kBit16 | kBit8,
+  kMaxBit8 = kBit8,
+
```
pgoodman: Visually, this is fun :-)

pgoodman: I'm in a valleyyyyyyy!

aaronyoo:
I am still working on the vector and array splitting

lib/Arch/Arch.cpp
aaronyoo: Fixed e40ab35

lib/Arch/Arch.cpp
aaronyoo: Fixed e40ab35

lib/Arch/Arch.cpp
aaronyoo: Handled the case of a struct split in e40ab35. However, I may need to do a minor bit of rearchitecting because arrays are split and vectors are split and packed.

lib/Arch/Arch.cpp
```diff
@@ -0,0 +1,294 @@
...
+  if (auto reg = TryRegisterAllocate(ty, reserved, constraints)) {
+    val_decl->reg = reg;
+    val_decl->type = &ty;
+  } else {
+    return nullptr;
+  }
+  return val_decl;
+}
+
+// +----------------------------------------------------------------------+
```
pgoodman: this is such a cool table

pgoodman: also, are you able to do a vector of i128? (including vectors of size 1)

lib/Arch/Arch.cpp
pgoodman: Is this resolved?

aaronyoo: Yes, fixed in 7db1b6f

lib/Arch/Arch.cpp
```diff
@@ -27,6 +29,54 @@ bool RegisterConstraint::ContainsVariant(const std::string& name) const {
...
+      } else {
+        LOG(FATAL) << "Unsupported (arch, os) pair: "
+                   << remill::GetArchName(arch->arch_name) << " "
+                   << remill::GetOSName(arch->os_name);
+      }
+      break;
+    }
+
+    // Fallthrough for unsupported architectures
+    case remill::kArchX86_AVX:
```
pgoodman: I think you should be able to handle the `_AVX` and `_AVX512` variants in the same way? Really, they will tell you whether or not you use `xmm*`, `ymm*`, or `zmm*`.

lib/Decl.cpp
```diff
@@ -286,31 +286,21 @@ llvm::json::Object ValueDecl::SerializeToJSON(const llvm::DataLayout &dl) {
...
   FunctionDecl decl;
   decl.type = func.getFunctionType();
-  
   decl.name = func.getName().data();
   decl.demangled_name = llvm::demangle(func.getName().data());
   decl.dl = &dl;
 
   // Try to guess the parameter and return value register allocation by looking
   // at the calling convention of the function
-  llvm::CallingConv::ID cc_id = func.getCallingConv();
```
pgoodman: A function might specify a calling convention, which might override the default. I think the `func.getCallingConv()` value should be passed in as a kind of "hint". If the value is `llvm::CallingConv::C` (this is the default value), then you do the arch+os-specifc heuristics, but if it's a non-default value, then use it.

lib/Arch/Arch.cpp
aaronyoo: Fixed in 455a2fc

lib/Arch/Arch.cpp
```diff
@@ -63,8 +64,6 @@ std::unique_ptr<CallingConvention> CallingConvention::CreateCCFromArch(
     }
 
     // Fallthrough for unsupported architectures
-    case remill::kArchX86_AVX:
-    case remill::kArchX86_AVX512:
     case remill::kArchAMD64_AVX:
```
pgoodman: The AVX variants of AMD64 can be moved up to the `case remill::kArchAMD64:` as well.

aaronyoo: Fixed in 7621822

lib/Decl.cpp
```diff
@@ -289,7 +289,9 @@ llvm::json::Object ValueDecl::SerializeToJSON(const llvm::DataLayout &dl) {
 FunctionDecl FunctionDecl::Create(const llvm::Function &func,
                                   const llvm::Module &mdl) {
   const llvm::DataLayout& dl = mdl.getDataLayout();
-  const remill::Arch *arch = remill::Arch::GetModuleArch(mdl);
+
+  remill::Arch::ArchPtr arch = remill::Arch::GetModuleArch(mdl);
```
pgoodman: Lets pass arch into this function, given that we're passing module into the function.

aaronyoo: Fixed e5d6087

lib/Arch/Arch.h
```diff
@@ -0,0 +1,269 @@
...
+                              bool injected_sret);
+
+ private:
+  // Register based parameter passing is generally not allowed for x86_C for
+  // types other than vector types. Even in the case of vector types it is
+  // important to note that if LLVM lowers something like a vector(2) of floats
+  // to <float, float> in IR, we will not be able to allocate it to a vector
+  // register because in our eyes it will no longer be a vector. This is
+  // consistent with the behavior of Clang but not GCC.
+  const std::vector<RegisterConstraint> parameter_register_constraints = {
```
pgoodman: Same _AVX and _AVX512 things.

lib/Arch/X86_C.cpp
```diff
@@ -0,0 +1,160 @@
...
+  switch (ret_type->getTypeID()) {
+    case llvm::Type::IntegerTyID:
+    case llvm::Type::PointerTyID: {
+      // Allocate EAX for an integer or pointer
+      value_declaration.reg = arch->RegisterByName("EAX");
+      break;
+    }
+    case llvm::Type::FloatTyID:
+    case llvm::Type::DoubleTyID:
+    case llvm::Type::FP128TyID: {
```
pgoodman: Here you want the x86 FP80 type; FP128 is for things like SPARC and maybe PPC.

lib/Decl.cpp
```diff
@@ -188,4 +195,131 @@ llvm::Value *FunctionDecl::CallFromLiftedBlock(
...
+  json.insert(
+      llvm::json::Object::KV{llvm::json::ObjectKey("parameters"),
+                             llvm::json::Value(std::move(params_json))});
+
+  llvm::json::Array returns_json;
+  for (auto rdecl : this->returns) {
+    returns_json.push_back(llvm::json::Value(rdecl.SerializeToJSON(*this->dl)));
+  }
+  json.insert(
+      llvm::json::Object::KV{llvm::json::ObjectKey("return values"),
```
pgoodman: `return_values`

lib/Decl.cpp
```diff
@@ -188,4 +195,131 @@ llvm::Value *FunctionDecl::CallFromLiftedBlock(
...
+    return_stack_pointer_json.insert(llvm::json::Object::KV{
+        llvm::json::ObjectKey("register"), this->return_stack_pointer->name});
+    return_stack_pointer_json.insert(llvm::json::Object::KV{
+        llvm::json::ObjectKey("offset"), this->return_stack_pointer_offset});
+    return_stack_pointer_json.insert(llvm::json::Object::KV{
+        llvm::json::ObjectKey("type"),
+        TranslateType(*this->return_stack_pointer->type, *this->dl)});
+
+    json.insert(llvm::json::Object::KV{
+        llvm::json::ObjectKey("return stack pointer"),
```
pgoodman: `return_stack_pointer`

lib/Decl.cpp
```diff
@@ -188,4 +195,131 @@ llvm::Value *FunctionDecl::CallFromLiftedBlock(
...
+        llvm::json::Value(std::move(return_stack_pointer_json))});
+  }
+  return json;
+}
+
+// Serialize a ParameterDecl to JSON
+llvm::json::Object ParameterDecl::SerializeToJSON(const llvm::DataLayout &dl) {
+  // Get the serialization for the ValueDecl
+  ValueDecl *val_decl_ptr = dynamic_cast<ValueDecl *>(this);
+  llvm::json::Object param_json = val_decl_ptr->SerializeToJSON(dl);
```
pgoodman: What you want is `this->ValueDecl::SerializeToJSON(dl)` in place of teh `dynamic_cast`.

lib/Decl.cpp
```diff
@@ -188,4 +195,131 @@ llvm::Value *FunctionDecl::CallFromLiftedBlock(
...
+
+  cc->AllocateSignature(decl, func);
+
+  // TODO(aty): for a better and more comprehensive serialization
+  // decl->address =
+  // decl->return_address =
+  // decl->return_stack_pointer =
+  // decl->return_stack_pointer_offset =
+  // decl->is_noreturn =
+  // decl->is_variadic =
```
pgoodman: `is_noreturn` and `is_variadic` are pretty easy to handle. You can do: `func.hasFnAttr(llvm::Attribute::NoReturn)` and probably `func.isVariadic()`, respectively.

PR #28, opened 2020-07-07
Fix x86 calling convention

kumarak:


lib/Arch/X86_C.cpp
```diff
@@ -203,6 +203,11 @@ llvm::Error X86_C::BindReturnValues(
 
 
   switch (ret_type->getTypeID()) {
+    case llvm::Type::VoidTyID:
+      LOG(WARNING) << "Could not allocate unsupported type" << remill::LLVMThingToString(ret_type).c_str()
+                   << " to return register in function " << function.getName().str().c_str();
+      return llvm::Error::success();
```
pgoodman, line 209: A void return type should be more than OK. Why the warning?

PR #32, opened 2020-08-10
Standalone compile without remill source checkout

ekilmer:
These CMake changes copy a lot of what's in remill, and I don't know if they're all necessary if we want to build anvill outside of the remill source tree, so I'm open to suggestions on a better approach. In particular, I'm wondering if we can get rid of the Z3-related changes, since Anvill doesn't use it.

Building outside of remill source is nice if you have it preinstalled already, and the CMake file(s) suggest that building outside of remill is possible, however it has proven a bit difficult without these changes.

I also encountered a compilation error with LLVM JSON handling and casting, so I fixed that, but I'm unsure if we actually want the fix how it is.


Dockerfile
```diff
@@ -1,49 +1,70 @@
...
+# Additional runtime dependencies go here
+FROM ${BUILD_BASE} as base
+ARG UBUNTU_VERSION
+RUN apt-get update && \
+    if [ "${UBUNTU_VERSION}" = "20.04" ] ; then \
+        apt-get install -qqy --no-install-recommends libtinfo6 ; \
+    else \
+        apt-get install -qqy --no-install-recommends libtinfo5 ; \
+    fi && \
+    apt-get install -qqy --no-install-recommends libz3-4 && \
```
artemdinaburg, line 20: So we should package z3 in trailofbits-libraries. We had some version specific issues before, such as the official version not being new enough or requiting OpenMP. 

If we *already* need trailofbits-libraries to build then lets not install this. If we *do not* need trailofbits-libraries, this may be a way to save some space in the docker image.

ekilmer: Thank you for the explanation. I just removed this and it should be a bit better now. I had forgotten to test without this before when I was running into other issues. 👍

ekilmer: The only thing in the Docker image should be Anvill and Remill (for the semantics files)

PR #40, opened 2020-08-18
Depend directly on remill

pgoodman:


Dockerfile
```diff
@@ -47,8 +47,8 @@ WORKDIR /anvill
 COPY . ./
 
 ENV PATH="${LIBRARIES}/llvm/bin:${LIBRARIES}/cmake/bin:${LIBRARIES}/protobuf/bin:${PATH}"
-ENV CC="${LIBRARIES}/llvm/bin/clang"
-ENV CXX="${LIBRARIES}/llvm/bin/clang++"
```
ekilmer, line 51: Why aren't we using the version of `clang` from `cxx-common`?

pgoodman: cxx-common is built with `/usr/bin/clang`, so I think it's best to have everything consistent.

ekilmer: Okay, that should be doable. I'll try to get this working in Docker

ekilmer: I pushed a commit to install `clang` during the `deps` stage of Docker, but it seems there are some other errors now regarding dependencies.

PR #47, opened 2020-09-28
Install python to venv

artemdinaburg:
* Install Anvill python to a virtualenv, if it exists
* Make Dockerfile create a virtualenv, so the Python API is installed in the `dist` container.


ekilmer (approved):
LGTM

PR #57, opened 2020-10-14
Add cache=False support for binja.py

pgoodman:


lib/Analyze.cpp
```diff
@@ -871,9 +883,9 @@ void RecoverMemoryAccesses(const Program &program, llvm::Module &module) {
   FindPossibleCrossReferences(program, module, "__anvill_sp", fixups,
                               maybe_fixups, sp_fixups);
 
-  LOG(WARNING) << "fixups.size()=" << fixups.size();
-  LOG(WARNING) << "maybe_fixups.size()=" << maybe_fixups.size();
-  LOG(WARNING) << "sp_fixups.size()=" << sp_fixups.size();
+  LOG(ERROR) << "fixups.size()=" << fixups.size();
```
surovic, line 886: Do we need to have this as an error? I think I demoted this to a warning because of CI.

pgoodman: No, it was for debugging.

artemdinaburg (approved):

PR #59, opened 2020-10-27
Updates anvill in what should be a backward-compatible way, ahead of updating any remill apis

pgoodman:
If this builds then we should merge it, as it should be backward-compatible with old remill, while also working with some newer remill changes.


lib/Arch/SPARC32_C.cpp
```diff
@@ -0,0 +1,347 @@
...
+  }
+
+  return llvm::createStringError(
+      std::errc::invalid_argument,
+      "Could not allocate unsupported type '%s' to return register in function '%s'",
+      remill::LLVMThingToString(ret_type).c_str(),
+      function.getName().str().c_str());
+}
+
+// For X86_64_SysV, the general argument passing behavior is, try to pass the
```
artemdinaburg, line 270: This comment needs to be updated to reflect SPARC things; its left over as copypasta from x86-64

lib/Arch/SPARC32_C.cpp
```diff
@@ -0,0 +1,347 @@
...
+      << "Injected struct returns are not supported on SPARC targets";
+
+  const auto param_names = TryRecoverParamNames(function);
+  llvm::DataLayout dl(function.getParent());
+
+  // Used to keep track of which registers have been allocated
+  AllocationState alloc_param(parameter_register_constraints, arch, this);
+  alloc_param.config.type_splitter = IntegerTypeSplitter;
+
+  uint64_t stack_offset = 92;
```
artemdinaburg, line 289: Transform to constant with meaningful name -- i am curious why this is 92

lib/Arch/SPARC32_C.cpp
```diff
@@ -0,0 +1,347 @@
...
+
+    auto allocation = alloc_param.TryRegisterAllocate(*param_type);
+
+    // Try to allocate from a register. If a register is not available then
+    // allocate from the stack.
+    if (allocation) {
+      const auto prev_size = parameter_declarations.size();
+
+      for (const auto &param_decl : allocation.getValue()) {
+        parameter_declarations.emplace_back();
```
artemdinaburg, line 305: any particular reason for using .emplace_back() and then .back() versus just making a new ParameterDecl? I figure its to keep the code generic using `auto` and type inference?

lib/Arch/SPARC32_C.cpp
```diff
@@ -0,0 +1,347 @@
...
+      if (bit_width <= 32) {
+        ret_values.emplace_back();
+        auto &value_declaration = ret_values.back();
+        value_declaration.reg = arch->RegisterByName("o0");
+        value_declaration.type = ret_type;
+        return llvm::Error::success();
+
+      // Split the integer across `o5:o0`. Experimentally, the largest
+      // returnable integer is 192 bits in size, any larger and LLVM crashes.
+      } else if (bit_width <= 192) {
```
artemdinaburg, line 195: Create a constant for 192

lib/Arch/SPARC64_C.cpp
```diff
@@ -0,0 +1,452 @@
...
+      if (bit_width <= 64) {
+        ret_values.emplace_back();
+        auto &value_declaration = ret_values.back();
+        value_declaration.reg = arch->RegisterByName("o0");
+        value_declaration.type = ret_type;
+        return llvm::Error::success();
+
+      // Split the integer across `o5:o0`. Experimentally, the largest
+      // returnable integer is 384 bits in size, any larger and LLVM crashes.
+      } else if (bit_width <= 384) {
```
artemdinaburg, line 289: Create a constant for 384

lib/Arch/SPARC32_C.cpp
```diff
@@ -0,0 +1,347 @@
...
+
+  llvm::Type *ret_type = function.getReturnType();
+  injected_sret = false;
+
+  // If there is an sret parameter then it is a special case. For the X86_C ABI,
+  // the sret parameters are guarenteed the be in %eax. In this case, we can
+  // assume the actual return value of the function will be the sret struct
+  // pointer.
+  if (function.hasStructRetAttr()) {
+    ret_values.emplace_back();
```
artemdinaburg, line 148: So in C++17 (which I think we have on? I can't confirm if we still need C++14 support) the emplace_back() call should return a reference to the newly created element. If we enforce 17, then we can get rid of all the .back() calls.

lib/Arch/SPARC32_C.cpp
pgoodman: it's used once, and is assigned to a name that is semi-meaningful. The source of the number escapes me; I think I did some experiments in compiler explorer and this number popped out. It's likely something to do with the size of a register window.

lib/Arch/SPARC32_C.cpp
pgoodman: I coded this before we adopted c++17, I think.

pgoodman: But also doing an emplace back, followed by a back, lets me operate on the decl in its final position.

lib/Arch/SPARC64_C.cpp
pgoodman: This isn't necessary imo; it's 6*64, i.e. the size of `o5:o4:o3:o2:o1:o0`.

Issue #72, opened 2020-12-16
Clang segfaults on Anvill-generated bitcode


artemdinaburg:
So it looks like Clang segfaults on bitcode generated by Anvill :). The current suspicion is that it is caused by inline assembly we generate to mark capture of registers not in the specification.

Proposing the following fixes

1) Gate the generation of inline assembly like:
```
  %9 = getelementptr inbounds %struct.State, %struct.State* %8, i32 0, i32 6, i32 1, i32 0, i32 0, !remill_register !5
  %10 = call i64 asm sideeffect "# read register RAX", "=r"()
  store i64 %10, i64* %9, align 8
  %11 = getelementptr inbounds %str
```
via a feature flag.

2) Implement a proper fix for capturing those registers. The ideal solution, from Slack:
```
really, the most accurate way to lift this would be two functions:
the first function would have the naked attribute; it'd contain only the inline assembly, read out the "missing" dependent registers, and pass them as explicit args to the second function, along with the already stated args
```

3) Submit a bug report to Clang since no matter how awful the bitcode, Clang shouldn't segfault.

PR #73, opened 2020-12-16
Gate inline assembly references of registers not in the spec

artemdinaburg:
Clang currently crashes on our use of inline assembly references. Guard the new behavior via a feature flag (see issue #72), and default to old behavior.

### DIFF
--- a/lib/Lift.cpp
+++ b/lib/Lift.cpp
@@ -212,27 +212,23 @@ static void DefineNativeToLiftedWrapper(const remill::Arch *arch,
   // the spec could feasibly miss some dependencies, and so after optimization,
   // we'll be able to observe uses of `__anvill_reg_*` globals, and handle
   // them appropriately.
-  if (FLAGS_feature_inline_asm_for_unspec_registers) {
 
-    arch->ForEachRegister([=, &ir](const remill::Register *reg_) {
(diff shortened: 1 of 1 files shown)


lib/Lift.cpp
```diff
@@ -206,19 +212,39 @@ static void DefineNativeToLiftedWrapper(const remill::Arch *arch,
   // the spec could feasibly miss some dependencies, and so after optimization,
   // we'll be able to observe uses of `__anvill_reg_*` globals, and handle
   // them appropriately.
-  arch->ForEachRegister([=, &ir](const remill::Register *reg_) {
-    if (auto reg = reg_->EnclosingRegister(); reg_ == reg) {
-      std::stringstream ss;
-      ss << "# read register " << reg->name;
+  if (FLAGS_feature_inline_asm_for_unspec_registers) {
```
pgoodman, line 215: Can you move this `if` statement to be inside the `arch->ForEachRegister`, following the definition of `ss`.

pgoodman (changes requested):

lib/Lift.cpp
artemdinaburg: Pushed updated version that moves the feature flag check inside the loop

pgoodman (approved):

PR #78, opened 2021-01-21
Enables more tests in CI

surovic:


ekilmer (approved):
LGTM! Thanks for updating the vcpkg CMake file also 👍

PR #91, opened 2021-02-04
Improvements to reference gathering via `refs_as_defs` in Binja frontend

surovic:
As the title suggests. The big thing here is that it introduces `--refs_as_defs` to `__main.py__` in the anvill module, which is **false** by default. This means that the script will **not** include definitions of functions or variables that the function specified by `--entry_point` uses. This also effectively disables recursive descent.

The `scripts/roundtrip.py` testing script takes this change into account, which is why CI works. However wider integration is likely to break.

### DIFF
--- a/python/anvill/binja.py
+++ b/python/anvill/binja.py
@@ -97,10 +97,19 @@ def is_undef(bv, inst):
 
 
 def is_code(bv, addr):
-    sec_list = bv.get_sections_at(addr)
-    for sec in sec_list:
+    for sec in bv.get_sections_at(addr):
         if sec.start <= addr < sec.end:
             return sec.semantics == bn.SectionSemantics.ReadOnlyCodeSectionSemantics
+    return False
+
+
+def is_data(bv, addr):
+    for sec in bv.get_sections_at(addr):
+        if sec.start <= addr < sec.end:
+            return sec.semantics in [
+                bn.SectionSemantics.ReadOnlyDataSectionSemantics,
+                bn.SectionSemantics.ReadWriteDataSectionSemantics,
+            ]
     return False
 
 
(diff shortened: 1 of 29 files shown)


python/anvill/binja.py
```diff
@@ -30,80 +30,89 @@ def is_valid_addr(bv, addr):
...
+    for sec in bv.get_sections_at(addr):
         if sec.start <= addr < sec.end:
             return sec.semantics == bn.SectionSemantics.ReadOnlyCodeSectionSemantics
     return False
 
 
+def is_data(bv, addr):
+    for sec in bv.get_sections_at(addr):
+        if sec.start <= addr < sec.end:
+            return sec.semantics in [
```
pgoodman, line 109: Recommend using `== ... or == ..`, or making a global variable that is a tuple that has the two semantic types in them, otherwise every time we do this check, a list of two elements is created only to be destroyed later.

python/anvill/binja.py
```diff
@@ -420,9 +456,9 @@ def _fill_bytes(self, bv, memory, start, end, ref_eas):
                 seg = bv.get_segment_at(ea)
                 br.seek(ea)
                 memory.map_byte(ea, br.read8(), seg.writable, seg.executable)
-                insn = self._bn_func.get_lifted_il_at(ea)
-                if insn:
-                    _collect_code_xrefs_from_insn(bv, insn, ref_eas)
+                inst = self._bn_func.get_low_level_il_at(ea)
+                if inst:
```
pgoodman, line 460: Missing a check of `and not is_unimplemented(inst):`?

pgoodman (changes requested):

pgoodman (approved):

PR #95, opened 2021-02-10
Handling of missing type class

kumarak:
Add warning logs for type classes that are not handled. Added checks if binja fails to get segment information for some variable address such as  `__elf_headers`.

### DIFF
--- a/python/anvill/var.py
+++ b/python/anvill/var.py
@@ -44,5 +44,7 @@ class Variable(object):
     def proto(self):
         proto = {}
         proto["address"] = self.address()
-        proto["type"] = self.type().proto(self._arch)
+        if self.type() != None:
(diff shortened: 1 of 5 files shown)


python/anvill/var.py
```diff
@@ -44,5 +44,6 @@ def is_declaration(self):
     def proto(self):
         proto = {}
         proto["address"] = self.address()
-        proto["type"] = self.type().proto(self._arch)
+        if self.type() != None:
+          proto["type"] = self.type().proto(self._arch)
```
pgoodman, line 48: 4-space indent

pgoodman (changes requested):

pgoodman (approved):

PR #97, opened 2021-02-11
anvill: Add StructType support to CreateConstFromMemory

alessandrogario:


lib/Lift.cpp
```diff
@@ -548,6 +548,30 @@ CreateConstFromMemory(const uint64_t addr, llvm::Type *type,
     case llvm::Type::PointerTyID: {
     } break;
 
+    case llvm::Type::StructTyID: {
```
artemdinaburg, line 551: Could you add a comment describing what this should do, at a high level? It looks like it finds structures and creates a static initializer for them?

alessandrogario: Done! I've added a comment inside the case

PR #98, opened 2021-02-11
New control flow support

sschriner:


include/anvill/MCToIRLifter.h
```diff
@@ -110,21 +110,38 @@ class MCToIRLifter {
...
                          llvm::BasicBlock *block);
+  void VisitConditionalIndirectJump(const remill::Instruction &inst,
+                           remill::Instruction *delayed_inst,
+                           llvm::BasicBlock *block);
+
   void VisitFunctionReturn(const remill::Instruction &inst,
                            remill::Instruction *delayed_inst,
                            llvm::BasicBlock *block);
 
+  void VisitConditionalFunctionReturn(const remill::Instruction &inst,
```
artemdinaburg, line 121: So I think I understand what a few of these should do, but the name, but what is a conditional function return?

lib/MCToIRLifter.cpp
```diff
@@ -123,13 +123,42 @@ void MCToIRLifter::VisitIndirectJump(const remill::Instruction &inst,
   remill::AddTerminatingTailCall(block, intrinsics.jump);
 }
 
+void MCToIRLifter::VisitConditionalIndirectJump(const remill::Instruction &inst,
```
artemdinaburg, line 126: Could you add a comment that describes at a high level what this visitor is doing? I see it makes a conditional branch, but I am curious why and what each side of the branch is logically doing

lib/MCToIRLifter.cpp
```diff
@@ -123,13 +123,42 @@ void MCToIRLifter::VisitIndirectJump(const remill::Instruction &inst,
...
+
+
 void MCToIRLifter::VisitFunctionReturn(const remill::Instruction &inst,
                                        remill::Instruction *delayed_inst,
                                        llvm::BasicBlock *block) {
   VisitDelayedInstruction(inst, delayed_inst, block, true);
   llvm::ReturnInst::Create(ctx, remill::LoadMemoryPointer(block), block);
 }
 
+void MCToIRLifter::VisitConditionalFunctionReturn(const remill::Instruction &inst,
```
artemdinaburg, line 150: Same as earlier, it would be really handy to have a comment about what this is doing, at a high level

lib/MCToIRLifter.cpp
```diff
@@ -217,6 +246,29 @@ void MCToIRLifter::VisitDirectFunctionCall(const remill::Instruction &inst,
   VisitAfterFunctionCall(inst, block);
 }
 
+void MCToIRLifter::VisitConditionalDirectFunctionCall(const remill::Instruction &inst,
```
artemdinaburg, line 249: Same as before regarding comments

lib/MCToIRLifter.cpp
```diff
@@ -226,6 +278,22 @@ void MCToIRLifter::VisitIndirectFunctionCall(const remill::Instruction &inst,
   VisitAfterFunctionCall(inst, block);
 }
 
+void MCToIRLifter::VisitConditionalIndirectFunctionCall(const remill::Instruction &inst,
```
artemdinaburg, line 281: Same as before regarding comments

artemdinaburg (commented):
The code looks sane from a quick look, but as someone who is not very familiar with Anvill internals yet, I would love some high-level comments about what each of the functions is supposed to be doing.

lib/MCToIRLifter.cpp
```diff
@@ -397,15 +465,27 @@ void MCToIRLifter::VisitInstruction(
...
       break;
+    case remill::Instruction::kCategoryConditionalDirectFunctionCall:
+      VisitConditionalDirectFunctionCall(inst, delayed_inst, block);
+      break;
     case remill::Instruction::kCategoryIndirectFunctionCall:
       VisitIndirectFunctionCall(inst, delayed_inst, block);
       break;
+    case remill::Instruction::kCategoryConditionalIndirectFunctionCall:
+      VisitConditionalIndirectFunctionCall(inst, delayed_inst, block);
+      break;
```
sschriner, line 488: These case statements will break building in the CI until the aarch32 branch is merged into Remill master. It builds locally for me using the aarch32 branch.

sschriner, line 488: This case statement will break the CI until the aarch32 remill branch is merged into master. It's building fine locally with that branch.

include/anvill/MCToIRLifter.h
pgoodman: It is a function return that is conditional on the condition codes :-P In aarch32, function calls/returns can be conditional.

lib/MCToIRLifter.cpp
```diff
@@ -123,13 +123,42 @@ void MCToIRLifter::VisitIndirectJump(const remill::Instruction &inst,
   remill::AddTerminatingTailCall(block, intrinsics.jump);
 }
 
+void MCToIRLifter::VisitConditionalIndirectJump(const remill::Instruction &inst,
+                         remill::Instruction *delayed_inst,
+                         llvm::BasicBlock *block) {
+  const auto lifted_func = block->getParent();
+  const auto cond = remill::LoadBranchTaken(block);
+  const auto taken_block = llvm::BasicBlock::Create(ctx, "", lifted_func);
+  remill::AddTerminatingTailCall(taken_block, intrinsics.jump);
```
pgoodman, line 132: I think this `AddTerminatingTailCall` probably should happen after doing the `VisitDelayedInstruction(inst, delayed_inst, taken_block, true);`

lib/MCToIRLifter.cpp
```diff
@@ -123,13 +123,42 @@ void MCToIRLifter::VisitIndirectJump(const remill::Instruction &inst,
...
+  const auto not_taken_block = llvm::BasicBlock::Create(ctx, "", lifted_func);
+  llvm::BranchInst::Create(taken_block, not_taken_block, cond, block);
+  VisitDelayedInstruction(inst, delayed_inst, taken_block, true);
+  VisitDelayedInstruction(inst, delayed_inst, not_taken_block, false);
+  llvm::BranchInst::Create(GetOrCreateBlock(inst.branch_taken_pc), taken_block);
```
pgoodman, line 137: A conditional indirect jump has no `inst.branch_taken_pc`, i.e. it is indirect, so it is not known, hence the reason for the terminating tail call to `__remill_jump`, so we don't need this. In place of this, you want the `AddTerminatingTailCall`.

lib/MCToIRLifter.cpp
pgoodman: As far as I can tell, handling of delayed instructions has also not been added here.

lib/MCToIRLifter.cpp
```diff
@@ -217,6 +246,29 @@ void MCToIRLifter::VisitDirectFunctionCall(const remill::Instruction &inst,
   VisitAfterFunctionCall(inst, block);
 }
 
+void MCToIRLifter::VisitConditionalDirectFunctionCall(const remill::Instruction &inst,
+                               remill::Instruction *delayed_inst,
+                               llvm::BasicBlock *block) {
+  const auto lifted_func = block->getParent();
+  VisitDelayedInstruction(inst, delayed_inst, block, true);
```
pgoodman, line 253: The `true` parameter here means: "lift the delayed instruction assuming the branch is taken", but I think you want to do both `true` and `false` delayed instruction handling here, as you had done with the indirect conditional branch[

lib/MCToIRLifter.cpp
pgoodman: Similar delayed instruction issues here.

lib/MCToIRLifter.cpp
```diff
@@ -226,6 +278,22 @@ void MCToIRLifter::VisitIndirectFunctionCall(const remill::Instruction &inst,
...
+                                 llvm::BasicBlock *block) {
+  const auto lifted_func = block->getParent();
+  VisitDelayedInstruction(inst, delayed_inst, block, true);
+  auto do_cond_call = llvm::BasicBlock::Create(ctx, "", lifted_func);
+  auto next_block = llvm::BasicBlock::Create(ctx, "", lifted_func);
+  llvm::BranchInst::Create(do_cond_call, next_block,
+                           remill::LoadBranchTaken(block), block);
+  remill::AddCall(block, intrinsics.function_call);
+  VisitAfterFunctionCall(inst, block);
+
```
pgoodman, line 292: Remove whitespace.

pgoodman (changes requested):

lib/MCToIRLifter.cpp
sschriner:
```
case Instruction::kCategoryConditionalDirectFunctionCall: {
          if (inst.branch_not_taken_pc == inst.branch_taken_pc) {
            goto direct_func_call;
          }
          try_add_delay_slot(true, block);
          trace_work_list.insert(inst.branch_taken_pc);
          auto target_trace = get_trace_decl(inst.branch_taken_pc);
          auto do_cond_call = llvm::BasicBlock::Create(context, "", func);
          auto next_block = GetOrCreateBranchNotTakenBlock();
          llvm::BranchInst::Create(do_cond_call, next_block,
                                   LoadBranchTaken(block), block);
          AddCall(do_cond_call, target_trace);

          const auto ret_pc_ref = LoadReturnProgramCounterRef(block);
          const auto next_pc_ref = LoadNextProgramCounterRef(block);
          llvm::IRBuilder<> ir(do_cond_call);
          ir.CreateStore(ir.CreateLoad(ret_pc_ref), next_pc_ref);
          ir.CreateBr(next_block);

          continue;
        }
```
I only add true based on there being only `try_add_delay_slot(true, block);` in what was added in the remill TraceLifter. Should there also be a `try_add_delay_slot` for `false` underneath the `try_add_delay_slot` for `true`?

pgoodman: There should be a `false` case becasue these are conditional calls; in an unconditional case, the call is always taken. The thing with delay slots on conditional control flow is that they have different "anulment" behaviour when the flow is taken vs. not-taken.

pgoodman: But yes, the tracelifter is probably also wrong.

PR #99, opened 2021-02-11
Handling of Named Type references

kumarak:
The PR fixes the handling of Named Type references in the binary. It recovers them as a block of bytes and lifts them. Recovering the elements of named types is still in the TODO list and will be handled in future PR.

### DIFF
--- a/lib/Lift.cpp
+++ b/lib/Lift.cpp
@@ -24,6 +24,8 @@
 #include <llvm/IR/Module.h>
 #include <llvm/Transforms/Scalar.h>
 #include <llvm/Transforms/Utils.h>
+
+#include <remill/BC/Compat/VectorType.h>
 #include <remill/BC/Util.h>
 
 #include <algorithm>
@@ -593,7 +595,7 @@ CreateConstFromMemory(const uint64_t addr, llvm::Type *type,
       result = llvm::ConstantArray::get(array_type, initializer_list);
     } break;
 
-    case llvm::Type::FixedVectorTyID: {
+    case llvm::GetFixedVectorTypeId(): {
       const auto vec_type = llvm::dyn_cast<llvm::FixedVectorType>(type);
       const auto num_elms = vec_type->getNumElements();
       const auto elm_type = vec_type->getElementType();
--- a/lib/Lift.cpp
+++ b/lib/Lift.cpp
@@ -564,7 +564,7 @@ CreateConstFromMemory(const uint64_t addr, llvm::Type *type,
       std::vector<llvm::Constant *> initializer_list;
       initializer_list.reserve(num_elms);
 
-      for (auto i = 0U; i < num_elms; ++i) {
+      for (auto i = 0u; i < num_elms; ++i) {
         const auto elm_type = struct_type->getStructElementType(i);
         const auto offset = layout->getElementOffset(i);
         auto const_elm = CreateConstFromMemory(addr + offset, elm_type, arch,
@@ -592,6 +592,23 @@ CreateConstFromMemory(const uint64_t addr, llvm::Type *type,
       }
       result = llvm::ConstantArray::get(array_type, initializer_list);
     } break;
+
+    case llvm::Type::FixedVectorTyID: {
+      const auto vec_type = llvm::dyn_cast<llvm::FixedVectorType>(type);
+      const auto num_elms = vec_type->getNumElements();
+      const auto elm_type = vec_type->getElementType();
+      const auto elm_size = dl.getTypeAllocSize(elm_type);
+      std::vector<llvm::Constant *> initializer_list;
+      initializer_list.reserve(num_elms);
+
+      for (auto i = 0u; i < num_elms; ++i) {
+        const auto elm_offset = i * elm_size;
+        auto const_elm = CreateConstFromMemory(addr + elm_offset, elm_type,
+                                               arch, program, module);
+        initializer_list.push_back(const_elm);
+      }
+      result = llvm::ConstantVector::get(initializer_list);
+    }break;
 
     default:
       LOG(FATAL) << "Unhandled LLVM Type: " << remill::LLVMThingToString(type);
--- a/lib/Lift.cpp
+++ b/lib/Lift.cpp
@@ -510,6 +510,7 @@ namespace {
 static llvm::APInt ReadValueFromMemory(const uint64_t addr, const uint64_t size,
                                        const remill::Arch *arch,
                                        const Program &program) {
+
   // create an instance of precision integer of size*8 bits
   llvm::APInt result(size * 8, 0);
   for (auto i = 0u; i < size; ++i) {
@@ -525,7 +526,7 @@ static llvm::APInt ReadValueFromMemory(const uint64_t addr, const uint64_t size,
   }
 
   // NOTE(artem): LLVM's APInt does not handle byteSwap()
-  // for size 8, leading to a segfault. Guard against it here.
+  // for size 1, leading to a segfault. Guard against it here.
   if (arch->MemoryAccessIsLittleEndian() && size > 1) {
     result = result.byteSwap();
   }
@@ -563,15 +564,14 @@ CreateConstFromMemory(const uint64_t addr, llvm::Type *type,
       std::vector<llvm::Constant *> initializer_list;
       initializer_list.reserve(num_elms);
 
-      for (std::uint64_t i = 0U; i < num_elms; ++i) {
+      for (auto i = 0U; i < num_elms; ++i) {
         const auto elm_type = struct_type->getStructElementType(i);
         const auto offset = layout->getElementOffset(i);
-        auto const_elm = CreateConstFromMemory(addr + offset, elm_type,
-                                               arch, program, module);
+        auto const_elm = CreateConstFromMemory(addr + offset, elm_type, arch,
+                                               program, module);
         initializer_list.push_back(const_elm);
       }
-      result = llvm::ConstantStruct::get(struct_type,
-                                         llvm::ArrayRef(initializer_list));
+      result = llvm::ConstantStruct::get(struct_type, initializer_list);
     } break;
 
     case llvm::Type::ArrayTyID: {
--- a/lib/Lift.cpp
+++ b/lib/Lift.cpp
@@ -510,8 +510,9 @@ namespace {
 static llvm::APInt ReadValueFromMemory(const uint64_t addr, const uint64_t size,
                                        const remill::Arch *arch,
                                        const Program &program) {
-  llvm::APInt result(size, 0);
-  for (auto i = 0u; i < (size / 8); ++i) {
+  // create an instance of precision integer of size*8 bits
+  llvm::APInt result(size * 8, 0);
+  for (auto i = 0u; i < size; ++i) {
     auto byte_val = program.FindByte(addr + i).Value();
     if (remill::IsError(byte_val)) {
       LOG(ERROR) << "Unable to read value of byte at " << std::hex << addr + i
@@ -525,7 +526,7 @@ static llvm::APInt ReadValueFromMemory(const uint64_t addr, const uint64_t size,
 
   // NOTE(artem): LLVM's APInt does not handle byteSwap()
   // for size 8, leading to a segfault. Guard against it here.
-  if (arch->MemoryAccessIsLittleEndian() && size > 8) {
+  if (arch->MemoryAccessIsLittleEndian() && size > 1) {
     result = result.byteSwap();
   }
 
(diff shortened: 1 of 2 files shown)


lib/Lift.cpp
```diff
@@ -450,7 +450,7 @@ llvm::Value *StoreNativeValue(llvm::Value *native_val, const ValueDecl &decl,
 
     return mem_ptr;
 
-  // Store it to memory.
+    // Store it to memory.
```
pgoodman, line 453: Please run the formatter script, found under `./scripts`, not clang-format directly.

python/anvill/binja.py
```diff
@@ -238,16 +238,52 @@ def _convert_bn_type(tinfo: bn.types.Type, cache):
...
+            ret = UnionType()
+            elem_type_ = ArrayType()
+            elem_type_.set_num_elements(tinfo.width)
+            ret.add_element_type(elem_type_)
+            cache[str(tinfo)] = ret
+            return ret
+
+        elif (named_tinfo.type_class
+            == bn.NamedTypeReferenceClass.TypedefNamedTypeClass):
+            ret = TypedefType()
```
pgoodman, line 266: I feel like you need to more here? Like, fill in the typedef name, or what it refers to?

python/anvill/binja.py
```diff
@@ -238,16 +238,52 @@ def _convert_bn_type(tinfo: bn.types.Type, cache):
...
+            cache[str(tinfo)] = ret
+            return ret
+
+        elif (named_tinfo.type_class
+            == bn.NamedTypeReferenceClass.EnumNamedTypeClass):
+            ret = EnumType()
```
pgoodman, line 272: Make sure to invoke `ret.set_underlying_type` correctly to an integer size based on the bit width of this enum.

python/anvill/binja.py
pgoodman: Probably using `ret.set_underlying_type`.

python/anvill/binja.py
```diff
@@ -238,16 +238,52 @@ def _convert_bn_type(tinfo: bn.types.Type, cache):
         width = tinfo.width
         return FloatingPointType(width)
 
+    elif tinfo.type_class == bn.TypeClass.NamedTypeReferenceClass:
+        named_tinfo = tinfo.named_type_reference
+        if (named_tinfo.type_class
+            == bn.NamedTypeReferenceClass.StructNamedTypeClass):
+            # Add array of size width as struct element type
+            # TODO(akshayk) How to recover the struct elements type??
```
pgoodman, line 246: This doesn't seem right; what's going on here? I think you want to call `add_element_type` multiple  times.

pgoodman (changes requested):

lib/Lift.cpp
```diff
@@ -569,6 +569,22 @@ CreateConstFromMemory(const uint64_t addr, llvm::Type *type,
       }
     } break;
 
+    case llvm::Type::StructTyID: {
```
artemdinaburg, line 572: Could you add a comment about what this is supposed to do, at a high level? It looks like it will statically initialize structure members?

python/anvill/binja.py
kumarak: I agree there is a need to do more with Typedef and Enum Named Types. I was not getting too many references about getting alias type of Named type references. Need to look more.

python/anvill/binja.py
kumarak: I could not find Binary ninja API's to get element members of struct Named type. So the elements are getting recovered as an array of size width.

I created an array [i8 x64] and added it as an element of the struct. The type will be recovered as ` "type": "=0{[Bx64]}"` and lifted as `%anvill.struct.0 = type { [64 x i8] }` 

The other solution could be adding the bytes type multiple time by calling the function `add_element_type`.

lib/Lift.cpp
kumarak: It statically initializes the global variables of type struct. Currently, it only supports the initialization for integer, pointer, or array types. I will update the comments. 
There is a PR from Alessandro with the same changes and it should be updated after the merge.

python/anvill/binja.py
kumarak: Set the Integer of size `tinfo.width` as the underlying type of an enum.

lib/Lift.cpp
```diff
@@ -546,57 +546,55 @@ CreateConstFromMemory(const uint64_t addr, llvm::Type *type,
...
+      std::vector<llvm::Constant *> initializer_list;
       auto elm_offset = 0;
 
-      std::vector<llvm::Constant *> const_list;
-
       for (std::uint64_t i = 0U; i < num_elms; ++i) {
-        auto elm_type = struct_type->getElementType(i);
-        auto elm_size = dl.getTypeSizeInBits(elm_type);
+        const auto elm_type = struct_type->getElementType(i);
+        auto elm_size = dl.getTypeAllocSize(elm_type);
```
pgoodman, line 566: Pulling structure elements out of memory is a bit more subtle than iterating over the structure types themselves. Can you change this code to use `dl.getStructLayout`, and then follow a similar API usage/structure as you see here: https://github.com/lifting-bits/remill/blob/master/lib/BC/Util.cpp#L1578-L1597

pgoodman (changes requested):

lib/Lift.cpp
kumarak: Updated it to use the `StructLayout` and use the struct type API's to get elements.

lib/Lift.cpp
```diff
@@ -546,53 +546,48 @@ CreateConstFromMemory(const uint64_t addr, llvm::Type *type,
     } break;
 
     case llvm::Type::PointerTyID: {
+      const auto pointer_type = llvm::dyn_cast<llvm::PointerType>(type);
+      const auto size = dl.getTypeSizeInBits(type);
+      auto val = ReadValueFromMemory(addr, size, arch, program);
+      result = llvm::Constant::getIntegerValue(pointer_type, val);
```
pgoodman, line 552: Woah, does `llvm::Constant::getIntegerValue` work on pointer types?

python/anvill/binja.py
```diff
@@ -173,8 +173,53 @@ def _convert_bn_llil_type(
...
+            ret.add_element_type(_convert_bn_type(bv, elem.type, cache))
+        return ret
+
+    elif (named_tinfo.type_class
+        == bn.NamedTypeReferenceClass.UnionNamedTypeClass):
+        # Get the union type and recover the member elements
+        ref_type = bv.get_type_by_name(named_tinfo.name);
+        struct_type =  ref_type.structure
+        ret = UnionType()
+        for elem in struct_type.union.members:
```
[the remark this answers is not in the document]
pgoodman: *before* adding the members.

python/anvill/binja.py
```diff
@@ -173,8 +173,53 @@ def _convert_bn_llil_type(
...
+        return
+
+    named_tinfo = tinfo.named_type_reference
+    if (named_tinfo.type_class
+        == bn.NamedTypeReferenceClass.StructNamedTypeClass):
+        # Get the bn struct type and recursively recover the elements
+        ref_type = bv.get_type_by_name(named_tinfo.name);
+        struct_type =  ref_type.structure
+        ret = StructureType()
+        for elem in struct_type.members:
```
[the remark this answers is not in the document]
pgoodman: *before* adding the members.

python/anvill/binja.py
```diff
@@ -238,16 +290,19 @@ def _convert_bn_type(tinfo: bn.types.Type, cache):
         width = tinfo.width
         return FloatingPointType(width)
 
+    elif tinfo.type_class == bn.TypeClass.NamedTypeReferenceClass:
+        ret = _convert_named_type_reference(bv, tinfo, cache)
+        cache[str(tinfo)] = ret
```
pgoodman, line 295: Remove cache line here. It should be done in each type.

python/anvill/binja.py
```diff
@@ -173,8 +173,53 @@ def _convert_bn_llil_type(
...
+        return ret
+
+    else:
+        DEBUG("WARNING: Unknown named type {} not handled".format(named_tinfo))
 
-def _convert_bn_type(tinfo: bn.types.Type, cache):
+
+def _convert_bn_type(bv, tinfo: bn.types.Type, cache):
     """Convert an bn `Type` instance into a `Type` instance."""
     if str(tinfo) in cache:
```
pgoodman, line 224: Can you add a function, `_cache_key`, that we will use in place of `str`, for converting a binary ninja type into a key for the cache, and implement as `str` for now, but so that it's easier to swap out later.

python/anvill/binja.py
```diff
@@ -266,7 +321,7 @@ def get_type(ty):
         return ty.type()
 
     elif isinstance(ty, bn.Type):
-        return _convert_bn_type(ty, {})
+        return _convert_bn_type(bv, ty, {})
```
pgoodman, line 324: Instead of using an empty dict here for the type cache, can you have `BNProgram` have a private, initially empty dictionary that is passed around? This means modifiying `get_type` to take in another parameter, or to take in the `BNProgram` and get at the `bv` through it.

pgoodman (changes requested):

lib/Lift.cpp
kumarak: It works on pointer types as well. It converts the APint value to pointers.

python/anvill/binja.py
kumarak: Thanks, @pgoodman for the suggestion. I would like to have this change in a separate PR. The changes are not many but It will be cleaner to have it that way. Let me know how do you think?

lib/Lift.cpp
```diff
@@ -540,59 +541,56 @@ CreateConstFromMemory(const uint64_t addr, llvm::Type *type,
...
+        const auto elm_type = struct_type->getStructElementType(i);
+        const auto offset = layout->getElementOffset(i);
+        auto const_elm = CreateConstFromMemory(addr + offset, elm_type,
+                                               arch, program, module);
+        initializer_list.push_back(const_elm);
       }
-
       result = llvm::ConstantStruct::get(struct_type,
-                                        llvm::ArrayRef(const_list));
+                                         llvm::ArrayRef(initializer_list));
```
surovic, line 574: Is the explicit conversion needed? IIRC there should be a conversion method from `std::vector` to `llvm::ArrayRef`.

lib/Lift.cpp
```diff
@@ -540,59 +541,56 @@ CreateConstFromMemory(const uint64_t addr, llvm::Type *type,
...
-      auto elm_offset = 0;
-
-      std::vector<llvm::Constant *> const_list;
+      const auto struct_type = llvm::dyn_cast<llvm::StructType>(type);
+      const auto layout = dl.getStructLayout(struct_type);
+      const auto num_elms = struct_type->getStructNumElements();
+      std::vector<llvm::Constant *> initializer_list;
+      initializer_list.reserve(num_elms);
 
       for (std::uint64_t i = 0U; i < num_elms; ++i) {
```
surovic, line 566: `auto` just for consistency with code around. Unless there's a reason to have a `uint64_t` here.

lib/Lift.cpp
```diff
@@ -540,59 +541,56 @@ CreateConstFromMemory(const uint64_t addr, llvm::Type *type,
...
-        result = llvm::ConstantDataArray::getString(module.getContext(), bytes,
-                                                    /*AddNull=*/false);
-      } else {
-        result = llvm::ConstantDataArray::getRaw(bytes, num_elms, elm_type);
+        const auto elm_offset = i * elm_size;
+        auto const_elm = CreateConstFromMemory(addr + elm_offset, elm_type,
+                                               arch, program, module);
+        initializer_list.push_back(const_elm);
       }
+      result = llvm::ConstantArray::get(array_type, initializer_list);
```
surovic, line 593: Will this produce the same output bitcode as before? Rellic migth rely on this to produce string literals later on.

surovic (changes requested):

lib/Lift.cpp
kumarak: The previous implementation was not handling the array of type structs. It should produce the same bitcode for string literals.

lib/Lift.cpp
```diff
@@ -510,8 +510,10 @@ namespace {
...
                                        const Program &program) {
-  llvm::APInt result(size, 0);
-  for (auto i = 0u; i < (size / 8); ++i) {
+
+  // create an instance of precision integer of size*8 bits
+  llvm::APInt result(size * 8, 0);
+  for (auto i = 0u; i < size; ++i) {
     auto byte_val = program.FindByte(addr + i).Value();
     if (remill::IsError(byte_val)) {
       LOG(ERROR) << "Unable to read value of byte at " << std::hex << addr + i
```
pgoodman, line 519: Reset the error log to `std::dec` after `(addr + i)`.

pgoodman: Otherwise all subsequent logs will log out line numbers in hex, which is unhelpful.

lib/Lift.cpp
```diff
@@ -540,59 +542,55 @@ CreateConstFromMemory(const uint64_t addr, llvm::Type *type,
   llvm::Constant *result{nullptr};
   switch (type->getTypeID()) {
     case llvm::Type::IntegerTyID: {
-      const auto size = dl.getTypeSizeInBits(type);
+      const auto size = dl.getTypeAllocSize(type);
       auto val = ReadValueFromMemory(addr, size, arch, program);
       result = llvm::ConstantInt::get(type, val);
     } break;
```
pgoodman, line 549: File an issue to handle float, double, half, x86_fp80, etc.

lib/Lift.cpp
```diff
@@ -540,59 +542,55 @@ CreateConstFromMemory(const uint64_t addr, llvm::Type *type,
...
+                                               program, module);
+        initializer_list.push_back(const_elm);
       }
-
-      result = llvm::ConstantStruct::get(struct_type,
-                                        llvm::ArrayRef(const_list));
+      result = llvm::ConstantStruct::get(struct_type, initializer_list);
     } break;
 
     case llvm::Type::ArrayTyID: {
```
pgoodman, line 577: Add vector types; they're basically a copy of arrays anyway.

python/anvill/binja.py
```diff
@@ -173,11 +173,64 @@ def _convert_bn_llil_type(
...
+
+    else:
+        DEBUG("WARNING: Unknown named type {} not handled".format(named_tinfo))
+
+
+def _convert_bn_type(bv, tinfo: bn.types.Type, cache):
     """Convert an bn `Type` instance into a `Type` instance."""
-    if str(tinfo) in cache:
-        return cache[str(tinfo)]
+    if _cache_key(tinfo) in cache:
```
pgoodman, line 232: Save `_cache_key(tinfo)` to a variable and re-use it everywhere in this function.

python/anvill/binja.py
pgoodman: I agree, in a separate PR, maybe make a class like `TypeLifter` or `TypeCache`, where the cache itself has methods to convert each binary ninja type, and internally manages a cache.

pgoodman (changes requested):

surovic (approved):

pgoodman (approved):

PR #100, opened 2021-02-11
Handling of variables with invalid address

kumarak:
The PR fixes the handling of variables with invalid addresses. It does not get added to the global variables list. Also, fixes the assertion error due to `unimplemented` or `undef` instructions as one of LLIL operand.

### DIFF
--- a/python/anvill/binja.py
+++ b/python/anvill/binja.py
@@ -131,8 +131,8 @@ def _collect_xrefs_from_inst(bv, inst, ref_eas, reftype=XrefType.XREF_NONE):
     if not isinstance(inst, bn.LowLevelILInstruction):
         return
 
-    assert not is_unimplemented(bv, inst)
-    assert not is_undef(bv, inst)
+    if is_unimplemented(bv, inst) or is_undef(bv, inst):
+        return
 
     if is_function_call(bv, inst) or is_jump(bv, inst):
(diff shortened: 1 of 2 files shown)


python/anvill/binja.py
```diff
@@ -131,8 +131,9 @@ def _collect_xrefs_from_inst(bv, inst, ref_eas, reftype=XrefType.XREF_NONE):
     if not isinstance(inst, bn.LowLevelILInstruction):
         return
 
-    assert not is_unimplemented(bv, inst)
-    assert not is_undef(bv, inst)
+    if (is_unimplemented(bv, inst)
```
pgoodman, line 134: get rid of extraneous `(` and `)` and move the `or ..` onto the same line as `is_unim...`

python/anvill/binja.py
```diff
@@ -514,6 +515,11 @@ def get_variable_impl(self, address):
         """Given an address, return a `Variable` instance, or
         raise an `InvalidVariableException` exception."""
 
+        # raise exception if the variable has invalid address
+        seg = self._bv.get_segment_at(address)
```
pgoodman, line 519: Add another (redundant) check on `get_segment_at` in the loop that tries to add referenced entities, so that we don't even enter here in the first place.

pgoodman (changes requested):

pgoodman (approved):

PR #102, opened 2021-02-12
Do not lift functions that are not in the JSON spec

surovic:
Modifies lifting to ignore functions that do not have mapped bytes in the JSON spec.

### DIFF
--- a/lib/Lift.cpp
+++ b/lib/Lift.cpp
@@ -590,21 +590,39 @@ bool LiftCodeIntoModule(const remill::Arch *arch, const Program &program,
 
   // Lift global variables.
   program.ForEachVariable([&](const anvill::GlobalVarDecl *decl) {
-    const auto addr = decl->address;
-    const auto name = anvill::CreateVariableName(addr);
-    const auto gvar = decl->DeclareInModule(name, module);
+    const auto addr{decl->address};
+    const auto name{anvill::CreateVariableName(addr)};
+    const auto gvar{decl->DeclareInModule(name, module)};
 
+    // Check if we have mapped bytes
+    if (!program.FindByte(addr)) {
+      return true;
+    }
     // Set initializer
-    auto init = CreateConstFromMemory(addr, decl->type, arch, program, module);
+    auto init{CreateConstFromMemory(addr, decl->type, arch, program, module)};
     gvar->setInitializer(init);
-
     return true;
   });
 
   // Lift functions.
   program.ForEachFunction([&](const FunctionDecl *decl) {
-    const auto entry = lifter.LiftFunction(*decl);
-    DefineNativeToLiftedWrapper(arch, *decl, entry);
+    // Initialize function entry
+    const auto name{CreateFunctionName(decl->address)};
+    FunctionEntry entry{nullptr, nullptr, nullptr};
+    entry.lifted_to_native =
+        remill::DeclareLiftedFunction(&module, name + ".lifted_to_native");
+    entry.native_to_lifted = decl->DeclareInModule(name, module, true);
+
+    // Check if we have mapped bytes
+    if (auto start{program.FindByte(decl->address)};
+        start && start.IsExecutable()) {
+      entry = lifter.LiftFunction(*decl);
+      DefineNativeToLiftedWrapper(arch, *decl, entry);
+    } else {
+      entry.native_to_lifted->removeFnAttr(llvm::Attribute::InlineHint);
+      entry.native_to_lifted->removeFnAttr(llvm::Attribute::AlwaysInline);
+      entry.native_to_lifted->addFnAttr(llvm::Attribute::NoInline);
+    }
     DefineLiftedToNativeWrapper(*decl, entry);
     OptimizeFunction(entry.native_to_lifted);
     return true;
--- a/lib/Lift.cpp
+++ b/lib/Lift.cpp
@@ -606,25 +606,28 @@ bool LiftCodeIntoModule(const remill::Arch *arch, const Program &program,
 
   // Lift functions.
   program.ForEachFunction([&](const FunctionDecl *decl) {
-    // Initialize function entry
-    const auto name{CreateFunctionName(decl->address)};
-    FunctionEntry entry{nullptr, nullptr, nullptr};
-    entry.lifted_to_native =
-        remill::DeclareLiftedFunction(&module, name + ".lifted_to_native");
-    entry.native_to_lifted = decl->DeclareInModule(name, module, true);
+    // Initialize function entry. This will lift machine code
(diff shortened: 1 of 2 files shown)


lib/Lift.cpp
```diff
@@ -590,21 +590,39 @@ bool LiftCodeIntoModule(const remill::Arch *arch, const Program &program,
...
   });
 
   // Lift functions.
   program.ForEachFunction([&](const FunctionDecl *decl) {
-    const auto entry = lifter.LiftFunction(*decl);
-    DefineNativeToLiftedWrapper(arch, *decl, entry);
+    // Initialize function entry
+    const auto name{CreateFunctionName(decl->address)};
+    FunctionEntry entry{nullptr, nullptr, nullptr};
+    entry.lifted_to_native =
```
artemdinaburg, line 612: What is the `lifted_to_native` function? I am assuming it's used when you pass this function's address to callbacks?

lib/Lift.cpp
```diff
@@ -590,21 +590,39 @@ bool LiftCodeIntoModule(const remill::Arch *arch, const Program &program,
...
+        remill::DeclareLiftedFunction(&module, name + ".lifted_to_native");
+    entry.native_to_lifted = decl->DeclareInModule(name, module, true);
+
+    // Check if we have mapped bytes
+    if (auto start{program.FindByte(decl->address)};
+        start && start.IsExecutable()) {
+      entry = lifter.LiftFunction(*decl);
+      DefineNativeToLiftedWrapper(arch, *decl, entry);
+    } else {
+      entry.native_to_lifted->removeFnAttr(llvm::Attribute::InlineHint);
```
artemdinaburg, line 622: Could you add a comment about what this is logically doing? It removes inlining attributes, but I can't immediately tell why and what high-level impact it has

lib/Lift.cpp
```diff
@@ -590,21 +590,39 @@ bool LiftCodeIntoModule(const remill::Arch *arch, const Program &program,
...
+    entry.lifted_to_native =
+        remill::DeclareLiftedFunction(&module, name + ".lifted_to_native");
+    entry.native_to_lifted = decl->DeclareInModule(name, module, true);
+
+    // Check if we have mapped bytes
+    if (auto start{program.FindByte(decl->address)};
+        start && start.IsExecutable()) {
+      entry = lifter.LiftFunction(*decl);
+      DefineNativeToLiftedWrapper(arch, *decl, entry);
+    } else {
```
artemdinaburg, line 621: Is it worthwhile to consider the case where data exists, but `false == start.IsExecutable()` ?

lib/Lift.cpp
```diff
@@ -590,21 +590,39 @@ bool LiftCodeIntoModule(const remill::Arch *arch, const Program &program,
...
+    const auto name{anvill::CreateVariableName(addr)};
+    const auto gvar{decl->DeclareInModule(name, module)};
 
+    // Check if we have mapped bytes
+    if (!program.FindByte(addr)) {
+      return true;
+    }
     // Set initializer
-    auto init = CreateConstFromMemory(addr, decl->type, arch, program, module);
+    auto init{CreateConstFromMemory(addr, decl->type, arch, program, module)};
```
artemdinaburg, line 602: Thanks for updating this, it looks like a solid C++14-ification of the prior code.

artemdinaburg (commented):
It looks good I would just like some more explanation about exactly what the change are doing in comments.

lib/Lift.cpp
surovic: Do the comments near [writeable](https://github.com/lifting-bits/anvill/blob/561e2f3ffda9b81de6878502946e3aa1791fd4bf/lib/Program.cpp#L70) and [executable](https://github.com/lifting-bits/anvill/blob/561e2f3ffda9b81de6878502946e3aa1791fd4bf/lib/Program.cpp#L77) byte metada answer the question?

IIRC this information is derived from the JSON spec so whatever the IDA or Binja frontend tells us is going to decide.

lib/Lift.cpp
pgoodman: You should look at the comments in `FunctionEntry`, they explain it well.

lib/Lift.cpp
pgoodman: I personally don't actually like this style for trivial types (where this is a pointer), but oh well. Assignment was easier to read/understand.

lib/Lift.cpp
```diff
@@ -590,21 +590,39 @@ bool LiftCodeIntoModule(const remill::Arch *arch, const Program &program,
...
     return true;
   });
 
   // Lift functions.
   program.ForEachFunction([&](const FunctionDecl *decl) {
-    const auto entry = lifter.LiftFunction(*decl);
-    DefineNativeToLiftedWrapper(arch, *decl, entry);
+    // Initialize function entry
+    const auto name{CreateFunctionName(decl->address)};
+    FunctionEntry entry{nullptr, nullptr, nullptr};
```
pgoodman, line 611: in this case, prefer: `FunctionEntry entry = {};` for zero-initialization.

lib/Lift.cpp
```diff
@@ -590,21 +590,39 @@ bool LiftCodeIntoModule(const remill::Arch *arch, const Program &program,
...
+    entry.lifted_to_native =
+        remill::DeclareLiftedFunction(&module, name + ".lifted_to_native");
+    entry.native_to_lifted = decl->DeclareInModule(name, module, true);
+
+    // Check if we have mapped bytes
+    if (auto start{program.FindByte(decl->address)};
+        start && start.IsExecutable()) {
+      entry = lifter.LiftFunction(*decl);
```
pgoodman, line 619: I don't understand why `LiftFunction` couldn't do the `FindByte` check and simply leave the function as a declaration instead. Even if we fill out additional entries that don't yet used, they don't matter, as they will be optimized away. Right now we have logic where there are two function entries, there is overriding, etc. and it's less obvious overall.

pgoodman (changes requested):

lib/Lift.cpp
```diff
@@ -590,23 +590,44 @@ bool LiftCodeIntoModule(const remill::Arch *arch, const Program &program,
...
+    // Wrap native functions in a function that lifted
+    // functions can call. This will result in the
+    // lifted functions calling the native ones.
     DefineLiftedToNativeWrapper(*decl, entry);
+
+    // Optimize and inline. After this we should end up
+    // with only native functions.
     OptimizeFunction(entry.native_to_lifted);
+
+    // The ritual is done.
```
pgoodman, line 630: 🤣

pgoodman (approved):

PR #104, opened 2021-02-12
Remove Python2.7 support

### LINKED ISSUE #86: Remove Python 2.x support
We're already adding type hints, and the Python 2.7 EOL should mean that we don't even bother trying to run a Python 2 setup.py.

surovic:
Resolves #86


ekilmer (approved):
LGTM

pgoodman (approved):

PR #111, opened 2021-02-17
CI: Update default CMake settings, fix package generation

alessandrogario:
Changes:
1. Enables the tests and the install target in the default configuration
2. Fixes an issues with packaging, which didn't work correctly due to how DESTDIR was handled


ekilmer (approved):
Changes LGTM. Could you provide maybe 1 sentence about what is being fixed in package generation with these changes.

alessandrogario:
> Changes LGTM. Could you provide maybe 1 sentence about what is being fixed in package generation with these changes.

I've updated the PR description with an explanation of the changes

PR #119, opened 2021-02-19
Refactors existing codebase to use Lifters and Providers

pgoodman:
…refactor while still using pre-existing code.

## Related issues

- [x] https://github.com/lifting-bits/anvill/issues/133 - anvill can isolate return addresses inside stack frames
- [ ] https://github.com/lifting-bits/anvill/issues/135 - anvill can lift raw memory into LLVM constants used to initialize global variables
- [x] https://github.com/lifting-bits/anvill/issues/134 - anvill can recover stack frames from the lifted functions
- [x] https://github.com/lifting-bits/anvill/issues/138 - anvill can fold/combine constant expressions through PHI and Select instructions


anvill/src/Lifters/Context.cpp
```diff
@@ -69,71 +69,54 @@ llvm::Constant *EntityLifterImpl::TryLiftData(
...
+// There is some collusion between the `EntityLifter` and the `FunctionLifter`
+// to ensure their view of the world remains consistent.
+void ContextImpl::SetLiftedFunction(uint64_t address, llvm::Function *func) {
+
+  address_to_entity[address] = func;
+  entity_to_address[func] = address;
+
+  // This function might call other functions, so make sure the entity lifter
+  // knows about the addresses of those other functions.
+  for (const auto &block : *func) {
```
surovic, line 82: Maybe consider [llvm::instructions(const llvm::Function &)](https://llvm.org/doxygen/InstIterator_8h.html). Alternatively maybe an [llvm::InstVisitor](https://llvm.org/doxygen/classllvm_1_1InstVisitor.html) if you anticipate this code will handle more instruction types, beside calls.

anvill/include/anvill/Lifters/Context.h
```diff
@@ -0,0 +1,78 @@
...
+class TypeProvider;
+class ValueLifter;
+
+// Lifting context for ANVILL. The lifting context keeps track of the options
+// used for lifting, the module into which lifted objects are placed, and
+// a the mapping between lifted objects and their original addresses in the
+// binary.
+class Context {
+ public:
+  ~Context(void);
```
surovic, line 47: Does this need to be declared explicitly?

anvill/include/anvill/Lifters/Context.h
```diff
@@ -0,0 +1,78 @@
...
+                   const std::shared_ptr<TypeProvider> &type_provider_);
+
+  // Assuming that `entity` is an entity that was lifted by this `EntityLifter`,
+  // then return the address of that entity in the binary being lifted.
+  std::optional<uint64_t> AddressOfEntity(llvm::GlobalValue *entity) const;
+
+  // Return the options being used by this entity lifter.
+  const LifterOptions &Options(void) const;
+
+  Context(const Context &) = default;
```
surovic, line 60: Same as with the destructor. Do we need to declare all the default constructors and assignment operators explicitly? Aside from the private one.

anvill/include/anvill/Lifters/DataLifter.h
```diff
@@ -0,0 +1,44 @@
...
+namespace anvill {
+
+class DataLifterImpl;
+
+class DataLifter {
+ public:
+  ~DataLifter(void);
+
+
+  DataLifter(const DataLifter &) = default;
```
surovic, line 35: Same as with `anvill::Context` above.

anvill/include/anvill/Lifters/FunctionLifter.h
```diff
@@ -0,0 +1,70 @@
...
+  // Returns an `llvm::Function *` that is part of `options_.module`.
+  //
+  // NOTE(pag): If this function returns `nullptr` then it means that we cannot
+  //            lift the function (e.g. bad address, or non-executable memory).
+  llvm::Function *LiftFunction(const FunctionDecl &decl) const;
+
+  // Declare the function associated with `decl` in the context's module.
+  llvm::Function *DeclareFunction(const FunctionDecl &decl) const;
+
+  FunctionLifter(const FunctionLifter &) = default;
```
surovic, line 55: Same as `anvill::Context` and `anvill::FunctionLifter`. If we don't need them, they shouldn't be here imo.

anvill/include/anvill/Lifters/ValueLifter.h
```diff
@@ -0,0 +1,69 @@
...
+
+  // Lift the bytes in `data` as an `llvm::Constant` into the module associated
+  // with `entity_lifter_`s options. This may produce an `llvm::Constant`
+  // that requires relocations, i.e. that depends on aliases or functions in
+  // `options_.module`, i.e. if we're lifting some data where the types contain
+  // pointer types, and thus we need to interpret bytes from `data` as being
+  // addresses and then lift them to pointers.
+  llvm::Constant *Lift(std::string_view data, llvm::Type *type_of_data) const;
+
+  ValueLifter(const ValueLifter &) = default;
```
surovic, line 52: You probably can guess by this point :)

anvill/src/Lifters/Context.cpp
```diff
@@ -0,0 +1,123 @@
...
+          if (called_func_addr) {
+            entity_to_address.emplace(called_func, *called_func_addr);
+          }
+        }
+      }
+    }
+  }
+}
+
+Context::~Context(void) {}
```
surovic, line 98: Same as with the default stuff in the headers.

anvill/src/Lifters/Context.h
```diff
@@ -0,0 +1,98 @@
...
+}  // namespace llvm
+namespace anvill {
+
+class ValueLifterImpl;
+
+// An entity lifter is responsible for lifting functions and data (variables)
+// into a target LLVM module.
+class ContextImpl {
+ public:
+  ~ContextImpl(void);
```
surovic, line 44: Same as above.

anvill/src/Lifters/FunctionLifter.cpp
```diff
@@ -0,0 +1,1389 @@
...
+FunctionLifterImpl::~FunctionLifterImpl(void) {}
+
+FunctionLifterImpl::FunctionLifterImpl(
+    const LifterOptions &options_, MemoryProvider &memory_provider_,
+    TypeProvider &type_provider_)
+    : options(options_),
+      memory_provider(memory_provider_),
+      type_provider(type_provider_),
+      semantics_module(remill::LoadArchSemantics(options.arch)),
+      context(semantics_module->getContext()),
```
surovic, line 118: Might be a good idea to rename this to reflect that this is an `LLVMContext`.

anvill/src/Lifters/FunctionLifter.cpp
```diff
@@ -0,0 +1,1389 @@
...
+          ->getType()};
+  auto type =
+      llvm::FunctionType::get(llvm::Type::getVoidTy(context), params, false);
+  return llvm::Function::Create(type, llvm::GlobalValue::ExternalLinkage, name,
+                                module);
+}
+
+}  // namespace
+
+FunctionLifterImpl::~FunctionLifterImpl(void) {}
```
surovic, line 109: Same as above.

anvill/src/Lifters/FunctionLifter.cpp
```diff
@@ -0,0 +1,1389 @@
...
+
+    switch (perms) {
+      case BytePermission::kUnknown:
+      case BytePermission::kReadableExecutable:
+      case BytePermission::kReadableWritableExecutable:
+        inst_out->bytes.push_back(static_cast<char>(byte));
+        break;
+      case BytePermission::kReadable:
+      case BytePermission::kReadableWritable:
+        goto found_all_bytes;
```
surovic, line 186: :eyes:

anvill/src/Lifters/FunctionLifter.cpp
```diff
@@ -0,0 +1,1389 @@
...
+  // Read the maximum number of bytes possible for instructions on this
+  // architecture. For x86(-64), this is 15 bytes, whereas for fixed-width
+  // architectures like AArch32/AArch64 and SPARC32/SPARC64, this is 4 bytes.
+  inst_out->bytes.reserve(max_inst_size);
+  for (auto i = 0u; i < max_inst_size; ++i) {
+    auto [byte, accessible, perms] = memory_provider.Query(addr + i);
+    switch (accessible) {
+      case ByteAvailability::kUnknown:
+      case ByteAvailability::kUnavailable:
+        goto found_all_bytes;
```
surovic, line 172: Same as below.

anvill/src/Providers/MemoryProvider.cpp
```diff
@@ -0,0 +1,76 @@
...
+#include <anvill/Program.h>
+
+namespace anvill {
+namespace {
+
+// Provider of memory wrapping around an `anvill::Program`.
+class ProgramMemoryProvider final : public MemoryProvider {
+ public:
+
+  virtual ~ProgramMemoryProvider(void) = default;
```
surovic, line 29: A `final` class with a virtual default destructor? As above, is this needed?

anvill/src/Providers/MemoryProvider.cpp
```diff
@@ -0,0 +1,76 @@
...
+
+ private:
+  ProgramMemoryProvider(void) = delete;
+
+  const Program &program;
+};
+
+}  // namespace
+
+MemoryProvider::~MemoryProvider(void) {}
```
surovic, line 68: Same as everywhere with the explicit destructors.

anvill/src/Providers/TypeProvider.cpp
```diff
@@ -0,0 +1,123 @@
...
+#include <glog/logging.h>
+
+namespace anvill {
+namespace {
+
+// Provider of memory wrapping around an `anvill::Program`.
+class ProgramTypeProvider final : public TypeProvider {
+ public:
+
+  virtual ~ProgramTypeProvider(void) = default;
```
surovic, line 39: Same as above.

anvill/src/Providers/TypeProvider.cpp
```diff
@@ -0,0 +1,123 @@
...
+  }
+
+  for (const auto &decl : types_it->second) {
+    typed_reg_cb(decl.reg->name, decl.type, decl.value);
+  }
+}
+
+}  // namespace
+
+TypeProvider::~TypeProvider(void) {}
```
surovic, line 104: Same as above.

anvill/src/Lifters/ValueLifter.h
```diff
@@ -0,0 +1,64 @@
...
+//            actually owns the memory (by value) of a `ValueLifterImpl`.
+//            There is a bit of a song and dance to pass around the right
+//            references, and this is why we see a reference to
+//            `EntityLifterImpl` passed along to `ValueLifterImpl::Lift`.
+class ValueLifterImpl {
+ public:
+  explicit ValueLifterImpl(const LifterOptions &options_);
+
+  // Consume `num_bytes` of bytes from `data`, and update `data` in place.
+  llvm::APInt ConsumeValue(std::string_view &data, unsigned num_bytes);
```
surovic, line 47: What does this return in the APInt?

anvill/src/Lifters/ValueLifter.h
```diff
@@ -0,0 +1,64 @@
...
+
+namespace llvm {
+class DataLayout;
+class LLVMContext;
+}  // namespace llvm
+namespace anvill {
+
+// Implementation of the `ValueLifter`.
+//
+// NOTE(pag): Due to the cyclic dependencies between `ValueLifter` and
```
surovic, line 35: There is no `EntityLifter` anymore.

anvill/src/Lifters/Context.h
```diff
@@ -0,0 +1,98 @@
...
+
+  // Provider of memory when asking for bytes for instructions or data.
+  const std::shared_ptr<MemoryProvider> memory_provider;
+
+  // Provider of type information when asking for function prototypes.
+  const std::shared_ptr<TypeProvider> type_provider;
+
+  // Lifts initializers of global variables.
+  //
+  // TODO(pag): This is a bit dumb, but there's a circular dependency here :-/
```
surovic, line 84: Is this circular dependency really needed? I haven't found any use of the `ValueLifter` inside `ContextImpl`. Maybe we won't need it?

anvill/src/Providers/MemoryProvider.cpp
pgoodman: No, typed it out. I don't use IDE templates / code generation.

anvill/src/Providers/MemoryProvider.cpp
pgoodman: Guarantees a sane vtable.

anvill/src/Lifters/Context.cpp
pgoodman: Oooh I was not aware of `llvm::instructions`!

anvill/src/Lifters/FunctionLifter.cpp
```diff
@@ -1537,20 +1538,18 @@ llvm::Function *FunctionLifter::AddFunctionToContext(
...
-      } else if (auto invoke = llvm::dyn_cast<llvm::InvokeInst>(&inst)) {
-        called_func = invoke->getCalledFunction();
-      }
-      if (called_func) {
-        const auto called_func_name = called_func->getName().str();
-        auto called_func_addr = AddressOfNamedFunction(called_func_name);
-        if (called_func_addr) {
-          lifter_context.AddEntity(called_func, *called_func_addr);
-        }
+  for (auto &inst : llvm::instructions(*func)) {
```
surovic, line 1541: We're now handling two different types of instructions. InstVisitor? If overhead is a concern:
>Note that this class is specifically designed as a template to avoid virtual function call overhead. Defining and using an InstVisitor is just as efficient as having your own switch statement over the instruction opcode. 
[Source](https://llvm.org/doxygen/classllvm_1_1InstVisitor.html)

surovic: Alternatively there's casting to a [llvm::CallBase](https://llvm.org/doxygen/classllvm_1_1CallBase.html)

pgoodman: Ahahaha performance on such a small thing isn't a concern :-P If it was I wouldn't create `std::string` here and there or heck, use LLVM :-P

Great suggestion on `CallBase`, made the change :-D

anvill/src/Providers/TypeProvider.cpp
pgoodman: This is basically the same thing as why llvm intorduces a bunch of dummy static variables here and there on classes using virtual inheritance; they want to make sure the vtables get produced and associated with a specific translation unit. I think explicit destructors is a pretty unobtrusive way about this.

anvill/src/Lifters/FunctionLifter.cpp
pgoodman: Done 👍

anvill/src/Lifters/FunctionLifter.cpp
pgoodman: Done 👍

anvill/src/Lifters/FunctionLifter.cpp
pgoodman: Renamed to `llvm_context`

anvill/include/anvill/Lifters/Context.h
pgoodman: Mostly I see it as a way of declaring what I intend to be valid behaviours. Omitting it means that you get default behaviour if it is possible, but it doesn't mean your API intends for the default behaviour to actually be used. Making it explicit communicates that you thought of how your classes can be constructed/copied, and declared to the world that the default implementation is in fact good. 

Really, we should use a macro to pump out these repetitive types of defaults, or repetitive deletes.

anvill/include/anvill/Lifters/Context.h
pgoodman: Insofar as I explicitly define the destructor, yes it does. To the extent that my defined destructor is empty and apparently useless, then no, it does not need to be declared or defined. This is a habit I have developed when using the pimpl pattern. I have this habit for two reasons:

 1. If we switched to a `unique_ptr` down the road in place of a `shared_ptr` then it'd be required to maintain our ability to forward declare but not define the implementation class. I often waffle on shared vs. unique, so this is the stronger of my two reasons.
 2. If we want non-trivial destruction in the future, then we can support it without ABI changes. This is a weaker point, because in practice we wouldn't put logic in `~Context`, but instead in `~ContextImpl`.

anvill/include/anvill/Lifters/Context.h
surovic: Honestly, if I had to pick between a macro and a repetitive piece of code, I'd take the code, since it's easier to read and I don't have to look for the macro definition or think about how the macro is going to unfold. As for communicating intent, the absence of assignment operators and copy constructors means that default behavior is "explicitly" intended, IMO. Hasn't rule-of-zero become sort of a standard?

anvill/include/anvill/Lifters/Context.h
surovic: Well, I'm not a huge fan of the pImpl pattern altogether due to adding an unnecessary layer of encapsulation / indirection between the public header and the implementation of the API. More often than not users of our code will end up reading the implementation due to an exhaustive documentation not being available and the pImpl pattern complicates this. That's just my opinion though, so maybe someone else can chime in on this.

libraries/anvill_passes/src/RecoverStackFrameInformation.cpp
```diff
@@ -0,0 +1,437 @@
...
+Result<std::monostate, StackAnalysisErrorCode>
+RecoverStackFrameInformation::UpdateFunction(
+    llvm::Function &function, const StackFrameAnalysis &stack_frame_analysis,
+    StackFrameStructureInitializationProcedure init_strategy,
+    std::size_t stack_frame_lower_padding,
+    std::size_t stack_frame_higher_padding) {
+
+  if (function.isDeclaration() ||
+      stack_frame_analysis.instruction_list.empty()) {
+    return StackAnalysisErrorCode::InvalidParameter;
```
kumarak, line 275: Here `instruction_list` could be empty for some function and it should not throw an error.  e.g:
```
define i32 @foo() #0 {
  %1 = call i32 @bar()
  ret i32 %1
}
```

pgoodman: Can you check if this shows up in Carson's branch?

PR #131, opened 2021-03-04
Implement new stack frame analysis

alessandrogario:
Issue and task details: https://github.com/lifting-bits/anvill/issues/134


libraries/anvill_passes/src/BaseFunctionPass.cpp
```diff
@@ -21,49 +21,16 @@
...
-  for (const auto &operand : operand_list) {
-    if (OperandReferencesStackPointer(operand)) {
-      return true;
-    }
-  }
-
-  return false;
-}
-
 bool BaseFunctionPass::InstructionReferencesStackPointer(
```
pgoodman, line 24: Update a cache mapping llvm::Value -> state, and introduce base cases to handle PHI nodes.

pgoodman: Key thing might be "when to clear the cache." You could hide the `runOnFunction` of an llvm pass as a derived virtual method, e.g. `virtual bool Run(llvm::Function &func) = 0;`, then your `BaseFunctionPass` implements a `runOnFunction` that clears out the cache, then returns `this->Run(func);`.

pgoodman: Then `BaseFunctionPass` would derive from `llvm::FunctionPass`.

pgoodman (changes requested):

anvill/include/anvill/Lifters/Options.h
```diff
@@ -144,6 +145,10 @@ class LifterOptions {
   // forwarding.
   bool store_inferred_register_values : 1;
 
+  // Instructs the RecoverStackFrameInformation function pass to zero-initialize
```
pgoodman, line 148: Eventually lets expand this to an enum that lets us not store, store undef, store zero, or fill with "symbolic" tracking values.

pgoodman (changes requested):

libraries/anvill_passes/src/RecoverStackFrameInformation.h
```diff
@@ -0,0 +1,114 @@
...
+// This structure contains the stack size, along with the lower and
+// higher bounds of the offsets, and all the `load` and `store` instructions
+// that have been analyzed
+struct StackFrameAnalysis final {
+  // Describes a `store` or `load` instruction that is operating on
+  // the stack, along with the SP-relative offset and the operand
+  // type
+  struct StackOperation final {
+    llvm::Instruction *instr{nullptr};
+    std::int32_t offset{};
```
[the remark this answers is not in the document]
alessandrogario: Related to https://github.com/lifting-bits/anvill/pull/131#discussion_r588736750

I have started casting offsets to std::int32_t since it seems like the std::int64_t displacement values returned by `TryResolveReference` are in fact 32-bits values that have not been sign-extended to 64-bit.

Should I try to convert to std::int64_t?

libraries/anvill_passes/src/RecoverStackFrameInformation.cpp
```diff
@@ -65,8 +65,9 @@ bool RecoverStackFrameInformation::Run(llvm::Function &function) {
...
   // analysis and use it to generate a stack frame type and update all the
   // store and load instructions
-  auto update_func_res = UpdateFunction(
-      function, stack_frame_analysis, options.zero_init_recovered_stack_frames);
+  auto update_func_res =
+      UpdateFunction(function, stack_frame_analysis,
+                     options.stack_frame_struct_init_procedure);
 
   if (!update_func_res.Succeeded()) {
     auto error_code =
```
pgoodman, line 73: If this is going to be frequent, then can you make an `ErrorString` or `ErrorCodeString` method that does this stringifying?

libraries/anvill_passes/src/RecoverStackFrameInformation.cpp
```diff
@@ -265,34 +266,32 @@ RecoverStackFrameInformation::UpdateFunction(
...
 
   // Pre-initialize the stack frame to zero if we have been requested
   // to do so. From the whole FunctionPass class, we get the setting
   // from the LiftingOptions class
-  if (initialize_stack_frame) {
-    for (auto i = 0U; i < stack_frame_analysis.size; ++i) {
-      auto stack_frame_byte = builder.CreateGEP(
-          stack_frame_alloca,
-          {builder.getInt32(0), builder.getInt32(0), builder.getInt32(i)});
+  if (init_strategy == StackFrameStructureInitializationProcedure::kZeroes) {
```
pgoodman, line 277: Use a `switch` on `init_strategy`, so adding new cases triggers a compiler warning/error for unhandled cases.

libraries/anvill_passes/tests/src/RecoverStackFrameInformation.cpp
```diff
@@ -35,11 +35,16 @@ namespace anvill {
 
 TEST_SUITE("RecoverStackFrameInformation") {
   TEST_CASE("Run the whole pass on aaa well-formed function") {
-    static const std::array<bool, 2> kZeroInitStackSettings = {false, true};
+    static const std::vector<StackFrameStructureInitializationProcedure>
```
pgoodman, line 38: You can also do: `static const auto kInitStackSettings[] = { ... };` and this will infer the enumeration type, and work in range-based for loops :-D

pgoodman (changes requested):

libraries/anvill_passes/src/BaseFunctionPass.h
```diff
@@ -123,6 +140,35 @@ bool BaseFunctionPass<UserFunctionPass>::IsMemoryOperation(
...
 }
 
+
+template <typename UserFunctionPass>
+Result<llvm::GlobalVariable *, BaseFunctionPassErrorCode>
+BaseFunctionPass<UserFunctionPass>::GetSymbolicValue(llvm::Module &module,
+                                                     llvm::Type *type,
+                                                     const std::string &name) {
+
+  static const std::string kMandatoryPrefix{"__anvill_"};
```
pgoodman, line 156: Add to ABI.h/.cpp

libraries/anvill_passes/src/RecoverStackFrameInformation.cpp
```diff
@@ -274,18 +300,51 @@ RecoverStackFrameInformation::UpdateFunction(
...
+
+    case StackFrameStructureInitializationProcedure::kSymbolic: {
+      auto current_offset = stack_frame_analysis.lowest_offset;
+      auto &module = *function.getParent();
+      auto byte_type = llvm::Type::getInt8Ty(module.getContext());
+
+      for (auto i = 0U; i < stack_frame_analysis.size; ++i) {
+        auto stack_frame_byte = builder.CreateGEP(
+            stack_frame_alloca,
+            {builder.getInt32(0), builder.getInt32(0), builder.getInt32(i)});
```
pgoodman, line 318: One trick I like to do here is make an array outside of the loop:
```c++
llvm::Value *indexes[] = {builder.getInt32(0), builder.getInt32(0), nullptr};
```

Then inside the loop, do:
```
indexes[2] = builder.getInt32(i);
```

And pass `indexes` to the `CreateGEP`.

libraries/anvill_passes/src/RecoverStackFrameInformation.cpp
```diff
@@ -274,18 +300,51 @@ RecoverStackFrameInformation::UpdateFunction(
...
+            GetStackSymbolicByteValue(module, current_offset);
+
+        if (!symbolic_value_res.Succeeded()) {
+          return symbolic_value_res.TakeError();
+        }
+
+        auto symbolic_value = symbolic_value_res.TakeValue();
+        ++current_offset;
+
+        auto casted_sym_value =
```
pgoodman, line 330: Ah, so here I think you can instead do an `ir.CreateLoad` to load "the symbolic byte", then an `ir.CreateStore` as you have to store it into the stack.

libraries/anvill_passes/src/RecoverStackFrameInformation.cpp
```diff
@@ -274,18 +300,51 @@ RecoverStackFrameInformation::UpdateFunction(
...
-             StackFrameStructureInitializationProcedure::kNone) {
-    return StackAnalysisErrorCode::InvalidParameter;
+    case StackFrameStructureInitializationProcedure::kUndef: {
+      auto null_value = llvm::UndefValue::get(stack_frame_type);
+      builder.CreateStore(null_value, stack_frame_alloca);
+
+      break;
+    }
+
+    case StackFrameStructureInitializationProcedure::kSymbolic: {
```
pgoodman, line 310: Can you add a comment here, describing what this is doing and why.

libraries/anvill_passes/src/RecoverStackFrameInformation.cpp
```diff
@@ -274,18 +300,51 @@ RecoverStackFrameInformation::UpdateFunction(
   // Pre-initialize the stack frame to zero if we have been requested
```
pgoodman, line 292: This comment is out of date. I think you want to move it down to before the `kZeroes` case.

libraries/anvill_passes/src/RecoverStackFrameInformation.cpp
```diff
@@ -274,18 +300,51 @@ RecoverStackFrameInformation::UpdateFunction(
...
+        auto casted_sym_value =
+            llvm::ConstantExpr::getPtrToInt(symbolic_value, byte_type);
+
+        builder.CreateStore(casted_sym_value, stack_frame_byte);
+      }
+
+      break;
+    }
+
+    case StackFrameStructureInitializationProcedure::kNone: break;
```
pgoodman, line 339: Comment that this may or may not be equivalent to storing undefined, and it depends on compiler switches and compiler version behaviour. Loaded values from uninitialized memory are likely implicitly undefined, whereas above, with `kUndef`, they load explicitly undefined.

libraries/anvill_passes/src/RecoverStackFrameInformation.cpp
```diff
@@ -274,18 +300,51 @@ RecoverStackFrameInformation::UpdateFunction(
...
   }
 
   // The stack analysis we have performed earlier contains all the
```
pgoodman, line 342: Can you add something after this along the lines of: if we're in a debug build, then run a new analysis, and it should fail to find to find any stack-related operations. This might help us detect unusual corner cases in the future.

libraries/anvill_passes/tests/src/RecoverStackFrameInformation.cpp
```diff
@@ -35,11 +35,12 @@ namespace anvill {
 
 TEST_SUITE("RecoverStackFrameInformation") {
   TEST_CASE("Run the whole pass on aaa well-formed function") {
```
pgoodman, line 37: "a"?

libraries/anvill_passes/tests/src/RecoverStackFrameInformation.cpp
```diff
@@ -61,6 +62,11 @@ TEST_SUITE("RecoverStackFrameInformation") {
         CHECK(RunFunctionPass(module.get(),
                               CreateRecoverStackFrameInformation(
                                   *error_manager.get(), lift_options)));
+
+        if (init_strategy ==
```
pgoodman, line 66: Can you comment on why we're printing the module out only for this strategy?

pgoodman (changes requested):

libraries/anvill_passes/src/RecoverStackFrameInformation.cpp
alessandrogario: I've added a "look at the enum definition" comment here and documented the strategies in more details in the public interface

libraries/anvill_passes/src/RecoverStackFrameInformation.cpp
alessandrogario: I've added a "look at the enum definition" comment here and documented the strategies in more details in the public interface

libraries/anvill_passes/tests/src/RecoverStackFrameInformation.cpp
alessandrogario: This was temporary code to debug the pass, and I have now removed it

libraries/anvill_passes/src/RecoverStackFrameInformation.cpp
alessandrogario: There is a test that covers this:

```c++
          // If we run a second stack analysis, we should no longer find any
          // stack frame operation to recover
          stack_frame_analysis_res = RecoverStackFrameInformation::AnalyzeStackFrame(function);
          REQUIRE(stack_frame_analysis_res.Succeeded());

          auto second_stack_frame_analysis = stack_frame_analysis_res.TakeValue();
          CHECK(second_stack_frame_analysis.stack_operation_list.empty());
```

libraries/anvill_passes/src/RecoverStackFrameInformation.h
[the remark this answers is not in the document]
pgoodman: Yeah, convert to `int64_t`, and then we should investigate `TryResolveReference` in a separate PR or Issue.

pgoodman (approved):

pgoodman (approved):

PR #140, opened 2021-03-19
passes: Add InstructionFolderPass

alessandrogario:


kumarak:
@alessandrogario, I think the type compare for the global variable will fail because it will be of pointer type but the type you are passing is `i8`.  Maybe you would like to compare it with the `symbolic_value->getValueType()`.  Sorry I was not able to add comments on the code because the changes are not from this PR.

https://github.com/lifting-bits/anvill/blob/alessandro/feature/implement-inst-combine-pass/libraries/anvill_passes/src/BaseFunctionPass.h#L183

libraries/anvill_passes/src/InstructionFolderPass.cpp
```diff
@@ -0,0 +1,632 @@
...
+    }
+  }
+
+  // Go through the visited instructions and drop the ones we no
+  // longer need. Doing the cleanup now ensures we do not risk
+  // new instructions polluting the unordered_set by ending up
+  // allocated at a memory location we have already seen
+  for (auto &instr : visited_instructions) {
+    if (instr->use_empty()) {
+      instr->eraseFromParent();
```
pgoodman, line 204: `function_changed = true;`

libraries/anvill_passes/src/InstructionFolderPass.cpp
```diff
@@ -0,0 +1,632 @@
...
+    // the replacement and the required cleanup
+    llvm::Instruction *replacement{nullptr};
+    if (!instr_folder(replacement, instr, condition, true_value, false_value,
+                      user_instr)) {
+      continue;
+    }
+
+    user_instr->replaceAllUsesWith(replacement);
+
+    replaced_instr_list.push_back(user_instr);
```
pgoodman, line 259: This is redundant; `replaceAllUsesWith` means there are no uses left, so you can do `replaced_instr->eraseFromParent();` here without the later loop and `use_empty` check.

pgoodman: If you want to be super safe (maybe there is some weird PHI node issue) then what I'd recommend is that `replace_instr_list` be a list a `std::pair`, and you `emplace_back(user_instr, replacement)`, then in your later loop, you do:

```c++
for (auto [instr, replacement_instr] : replaced_instr_list) {
  instr->replaceAllUsesWith(replacement_instr);
  instr->eraseFromParent();
}
```

libraries/anvill_passes/src/InstructionFolderPass.cpp
```diff
@@ -0,0 +1,632 @@
...
+    const auto &instr_folder = instr_folder_it->second;
+
+    // If we succeed in folding this instruction, then also perform
+    // the replacement and the required cleanup
+    llvm::Instruction *replacement{nullptr};
+    if (!instr_folder(replacement, instr, incoming_value_list, user_instr)) {
+      continue;
+    }
+
+    user_instr->replaceAllUsesWith(replacement);
```
pgoodman, line 319: Ditto on the above comment.

libraries/anvill_passes/src/InstructionFolderPass.cpp
```diff
@@ -0,0 +1,632 @@
...
+  // other operand needs to be a constant
+  if (!llvm::isa<llvm::Constant>(operand)) {
+    return false;
+  }
+
+  // Finally, rewrite the instructions
+  llvm::IRBuilder<> builder(binary_op_instr);
+
+  auto opcode =
+      llvm::dyn_cast<llvm::BinaryOperator>(binary_op_instr)->getOpcode();
```
pgoodman, line 420: Just use `binary_op_instr->getOpcode()`

pgoodman (changes requested):

alessandrogario:
> @alessandrogario, I think the type compare for the global variable will fail because it will be of pointer type but the type you are passing is `i8`. Maybe you would like to compare it with the `symbolic_value->getValueType()`. Sorry I was not able to add comments on the code because the changes are not from this PR.
> 
> https://github.com/lifting-bits/anvill/blob/alessandro/feature/implement-inst-combine-pass/libraries/anvill_passes/src/BaseFunctionPass.h#L183

Thanks for the review! I will start a new branch to fix it and look at an issue i found with the undefined behaviour sanitizer in the stack recovery function pass

pgoodman (approved):

pgoodman:
> > @alessandrogario, I think the type compare for the global variable will fail because it will be of pointer type but the type you are passing is `i8`. Maybe you would like to compare it with the `symbolic_value->getValueType()`. Sorry I was not able to add comments on the code because the changes are not from this PR.
> > https://github.com/lifting-bits/anvill/blob/alessandro/feature/implement-inst-combine-pass/libraries/anvill_passes/src/BaseFunctionPass.h#L183
> 
> Thanks for the review! I will start a new branch to fix it and look at an issue i found with the undefined behaviour sanitizer in the stack recovery function pass

Ah indeed, that does need to a `getValueType` or maybe `getElementType`. There's a Remill API for this :-P https://github.com/lifting-bits/remill/blob/3808e9951db8a7332f361d802de5815cb2877571/include/remill/BC/Compat/GlobalValue.h#L24

PR #149, opened 2021-03-30
Implement control flow redirection

alessandrogario:


anvill/include/anvill/Lifters/EntityLifter.h
```diff
@@ -54,7 +55,8 @@ class EntityLifter {
   explicit EntityLifter(
       const LifterOptions &options,
       const std::shared_ptr<MemoryProvider> &mem_provider_,
-      const std::shared_ptr<::anvill::TypeProvider> &type_provider_);
+      const std::shared_ptr<::anvill::TypeProvider> &type_provider_,
+      const Program &program);
```
pgoodman, line 59: Pass through the control-flow provider instead; we eventually want to get rid of or factor out `Program`.

pgoodman: Perhaps instead we should put the providers into lifter options? What do you think? I.e. one needs to do `LifterOptions::Create(providers here)` to get back a value type, then you can fiddle with the options. Another possibility would be a lifter context which has a lifter options and each provider.

anvill/include/anvill/Program.h
```diff
@@ -239,6 +239,14 @@ class Program {
       const std::string &name,
       std::function<bool(const FunctionDecl *)> callback) const;
 
+  // Returns a possible control flow redirection for the given address
+  // or the input address itself if nothing is found
+  bool GetControlFlowRedirection(std::uint64_t &destination,
```
pgoodman, line 244: `TryGet...` I like the `bool`-returning things to have `Try` prefix to convey in the name that it can fail.

anvill/python/anvill/binja/bnprogram.py
```diff
@@ -0,0 +1,275 @@
...
+                        function_thunk.name == "__libc_start_main"
+                        or function_thunk.name not in caller_function.name
+                    ):
+                        continue
+
+                    redirection_source = caller_function.start
+
+                    print(
+                        "anvill: Redirecting thunk {:x}/{} user {:x} to {:x}".format(
+                            function_thunk.rva,
```
pgoodman, line 219: I noticed this `.rva` use; elsewhere I thing we use things like `.start`; should we be switching to `.rva`?

anvill/src/Lifters/FunctionLifter.cpp
```diff
@@ -317,10 +327,24 @@ void FunctionLifter::VisitConditionalFunctionReturn(
...
+  llvm::BranchInst::Create(GetOrCreateTargetBlock(inst.branch_not_taken_pc),
                            not_taken_block);
 }
 
+std::optional<FunctionDecl> FunctionLifter::TryGetTargetFunctionType(std::uint64_t address) {
+  auto redirected_addr = control_flow_provider->GetRedirection(address);
+
+  auto opt_function_decl = type_provider.TryGetFunctionType(redirected_addr);
+  if (!opt_function_decl.has_value()) {
+    return std::nullopt;
```
pgoodman, line 339: Don't fail here, fall through to try to get the `function_decl` for the original address.

pgoodman: So the logic should be:

```
if (!opt_function_decl.has_value()) {
  if (address != redirected_address) {
    opt_function_decl = type_provider.TryGetFunctionType(address);
    if (opt_function_decl.has_value()) {
      opt_function_decl->address = redirected_addr;
    }
  }
}
return opt_function_decl;
```

anvill/src/Lifters/FunctionLifter.cpp
```diff
@@ -988,18 +1013,20 @@ void FunctionLifter::VisitInstructions(uint64_t address) {
...
         if (const auto mem_ptr_from_call =
-                TryCallNativeFunction(inst_addr, other_decl, block)) {
+                TryCallNativeFunction(decl.address, other_decl, block)) {
           llvm::ReturnInst::Create(llvm_context, mem_ptr_from_call, block);
           continue;
         }
 
         LOG(ERROR) << "Failed to call native function at " << std::hex
-                   << inst_addr
+                   << decl.address
```
pgoodman, line 1029: Change to: `"Failed to call native function " << std::hex << decl.address << " at " << inst_addr << " via ...`

anvill/src/Lifters/FunctionLifter.cpp
```diff
@@ -1509,7 +1536,7 @@ llvm::Function *FunctionLifter::LiftFunction(const FunctionDecl &decl) {
   // instruction.
   //
   // NOTE(pag): This also introduces the first element to the work list.
-  ir.CreateBr(GetOrCreateBlock(func_address));
+  ir.CreateBr(GetOrCreateTargetBlock(func_address));
```
pgoodman, line 1539: Use `GetOrCreateBlock` instead of `GetOrCreateTargetBlock`. If we're purposefully lifting a thunk, then I think we want to lift it and not redirect it. Leave a comment about that here as a future TODO to think about.

pgoodman (changes requested):

anvill/src/Lifters/FunctionLifter.cpp
```diff
@@ -317,10 +327,31 @@ void FunctionLifter::VisitConditionalFunctionReturn(
...
 }
 
+std::optional<FunctionDecl> FunctionLifter::TryGetTargetFunctionType(std::uint64_t address) {
+  auto redirected_addr = control_flow_provider->GetRedirection(address);
+
+  // In case we get redirected but still fail, try once more with the original
+  // address
+  auto opt_function_decl = type_provider.TryGetFunctionType(redirected_addr);
+  if (!opt_function_decl.has_value() && redirected_addr != address) {
+    redirected_addr = address;
```
pgoodman, line 341: We don't want to undo the `redirected_addr` here.

pgoodman (changes requested):

anvill/include/anvill/Lifters/Options.h
```diff
@@ -106,6 +110,9 @@ class LifterOptions {
   // Target module into which code will be lifted.
   llvm::Module *const module;
 
+  // The control flow provider, used for thunk redirections
```
pgoodman, line 113: More elaborate comment needed, and also use periods in your sentences. The control-flow provider is used for thunk redirection (which helps to produce nicer bitcode for external function calls). It also provides possible targets for indirect jumps that are represented by jump tables.

pgoodman (approved):

PR #158, opened 2021-04-09
Fix intrinsic for return instructions lifted as indirect jump

kumarak:
Added a pass that identifies if a return instruction is lifted as an indirect jump and transforms the remill intrinsic calls for them.

### DIFF
--- a/libraries/anvill_passes/src/TransformRemillJumpIntrinsics.cpp
+++ b/libraries/anvill_passes/src/TransformRemillJumpIntrinsics.cpp
@@ -22,15 +22,11 @@
 #include <glog/logging.h>
 #include <llvm/ADT/Triple.h>
 #include <llvm/IR/Function.h>
-#include <llvm/IR/IRBuilder.h>
 #include <llvm/IR/InstIterator.h>
 #include <llvm/IR/Instruction.h>
-#include <llvm/IR/Instructions.h>
 #include <llvm/IR/Intrinsics.h>
-#include <llvm/IR/Metadata.h>
 #include <llvm/IR/Module.h>
 #include <llvm/Pass.h>
-#include <remill/BC/ABI.h>
 #include <remill/BC/Util.h>
 
 #include <utility>
--- a/libraries/anvill_passes/src/TransformRemillJumpIntrinsics.cpp
+++ b/libraries/anvill_passes/src/TransformRemillJumpIntrinsics.cpp
@@ -111,7 +111,7 @@ TransformRemillJumpIntrinsics::QueryReturnAddress(const llvm::DataLayout &dl,
 // with the given function type
 static llvm::Function *FindIntrinsic(llvm::Module *module,
                                      llvm::FunctionType *type,
-                                     std::string name) {
+                                     const std::string &name) {
   auto function = module->getFunction(name);
   if (!function) {
     function = llvm::Function::Create(type, llvm::GlobalValue::ExternalLinkage,
--- a/libraries/anvill_passes/src/TransformRemillJumpIntrinsics.cpp
+++ b/libraries/anvill_passes/src/TransformRemillJumpIntrinsics.cpp
@@ -1,0 +1,233 @@
+/*
+ * Copyright (c) 2021 Trail of Bits, Inc.
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as
+ * published by the Free Software Foundation, either version 3 of the
+ * License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Affero General Public License for more details.
+ *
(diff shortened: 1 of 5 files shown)


libraries/anvill_passes/src/TransformRemillJumpIntrinsics.cpp
```diff
@@ -0,0 +1,229 @@
...
+
+char TransformRemillJumpIntrinsics::ID = '\0';
+
+// Returns `true` if `val` is a possible return address
+ReturnAddressResult
+TransformRemillJumpIntrinsics::QueryReturnAddress(const llvm::DataLayout &dl,
+                                                  llvm::Value *val) const {
+
+  if (auto call = llvm::dyn_cast<llvm::CallBase>(val)) {
+    if (call->getIntrinsicID() == llvm::Intrinsic::returnaddress) {
```
pgoodman, line 83: I think there is an analysis function like `IsRelatedToReturnAddress` or `IsReturnAddress`

libraries/anvill_passes/src/TransformRemillJumpIntrinsics.cpp
```diff
@@ -0,0 +1,229 @@
...
+      return kReturnAddressProgramCounter;
+    } else if (auto func = call->getCalledFunction()) {
```
pgoodman, line 85: This is an interesting case that may be missing from the aforementioned functions, i.e. a remill memory read (and missing, a load inst case. Perhaps update the analysis function with these cases you've marked.

libraries/anvill_passes/src/TransformRemillJumpIntrinsics.cpp
```diff
@@ -0,0 +1,229 @@
...
+  function->addFnAttr(llvm::Attribute::NoInline);
+  function->removeFnAttr(llvm::Attribute::NoReturn);
+  function->removeFnAttr(llvm::Attribute::UWTable);
+  function->removeFnAttr(llvm::Attribute::AlwaysInline);
+  return function;
+}
+
+// Find the call site of the given function and add them to vector
+// if `pred(call)` is true
+static std::vector<llvm::CallBase *>
```
pgoodman, line 144: Given that this is `static`, instead of using `std::function`, use a template typename, so that the compiler can inline the predicate lambda into the static function. `std::function` often hurts optimization.

libraries/anvill_passes/src/TransformRemillJumpIntrinsics.cpp
```diff
@@ -0,0 +1,229 @@
...
+#include <remill/BC/Util.h>
+
+#include <utility>
+#include <vector>
+
+
+namespace anvill {
+namespace {
+
+const std::string_view remill_jump = "__remill_jump";
```
pgoodman, line 39: Make these into `llvm::StringRef` so that below you don't need to `== remill_jump.data()`

libraries/anvill_passes/src/TransformRemillJumpIntrinsics.cpp
```diff
@@ -0,0 +1,229 @@
...
+  }
+
+  return kUnclassifiableProgramCounter;
+}
+
+// Find the remill intrinsic in module. If it is missing create one
+// with the given function type
+static llvm::Function *FindIntrinsic(llvm::Module *module,
+                                     llvm::FunctionType *type,
+                                     const char *name) {
```
pgoodman, line 124: Make this take an `llvm::StringRef` as the name.

libraries/anvill_passes/src/TransformRemillJumpIntrinsics.cpp
```diff
@@ -0,0 +1,229 @@
...
+
+
+// Try to identify the patterns of `__remill_function_call` that we can
+// remove.
+bool TransformRemillJumpIntrinsics::runOnFunction(llvm::Function &func) {
+  const auto module = func.getParent();
+  const auto &dl = module->getDataLayout();
+  auto calls = FindFunctionCalls(func, [&](llvm::CallBase *call) -> bool {
+    const auto func = call->getCalledFunction();
+    if (!func || !(func->getName() == remill_jump.data())) {
```
pgoodman, line 189: `!func || func->getName() != remill_jump`

libraries/anvill_passes/src/TransformRemillJumpIntrinsics.cpp
pgoodman, line 39: Rename to `kRemillJumpIntrinsicName`. Ditto for call.

pgoodman (changes requested):

libraries/anvill_passes/src/TransformRemillJumpIntrinsics.cpp
```diff
@@ -121,7 +111,7 @@ TransformRemillJumpIntrinsics::QueryReturnAddress(const llvm::DataLayout &dl,
 // with the given function type
 static llvm::Function *FindIntrinsic(llvm::Module *module,
                                      llvm::FunctionType *type,
-                                     const char *name) {
+                                     std::string name) {
```
pgoodman, line 114: const ref

pgoodman (approved):
One tiny change, but this looks 👍

PR #159, opened 2021-04-11
Add test for transform remill jump pass

kumarak:
Add test for `TranformRemillJumpIntrinsics` pass. The llvm IR is generated from the instructions using a jump to return from the function.

```
   0:  mov    $0x8,%eax
   5:  pop    %rbx
   6:  jmpq   *%rbx
```


libraries/anvill_passes/tests/src/TransformRemillJump.cpp
```diff
@@ -0,0 +1,53 @@
...
+
+    anvill::LifterOptions options(arch.get(), *module.get(), nullptr);
+
+    // memory and types will not get used and create lifter with null
+    anvill::EntityLifter lifter(options, nullptr, nullptr);
+
+    CHECK(RunFunctionPass(module.get(),
+                          CreateTransformRemillJumpIntrinsics(lifter)));
+
+    REQUIRE(module->getFunction("__remill_jump") == nullptr);
```
pgoodman, line 49: I think what you should do is:
Require that the function `__remill_function_return` can be found and has a single user
Require that the function `__remill_jump` cannot be found, and if it can be, that is has no users.

I.e.

```c++
const auto ret_func = module->getFunction("__remill_function_return");
const auto jump_func = module->getFunction("__remill_jump");

REQUIRE(ret_func && !ret_func->use_empty());
REQUIRE(!jump_func || jump_func->use_empty());
```

pgoodman (changes requested):
Can you also add a that does:

```
add rsp, 8
mov eax, 8
jmp qword ptr [rsp]
```

As well as some aarch32 tests with various behaviours which you've identified as problematic?

kumarak:
@pgoodman, I added a few more generated bitcode from amd64 and aarch32 binaries. It includes the instruction set you mentioned. 
```   
   0:	str	lr, [sp]
   4:	mov	r0, #0
   8:	ldr	r2, [sp]
   c:	mov	pc, r2
```
```
   0:	push	{r2, lr}
   4:	cmp	r1, #1
   8:	moveq	r0, r1
   c:	movne	r0, #0
  10:	pop	{r2, pc}
```

pgoodman (approved):

Issue #150, opened 2021-03-30
InstructionFolderPass: Crash when folding PHI nodes and erasing unused incoming values


alessandrogario:
Test case: [anvill_issue_150_test_case.json.zip](https://github.com/lifting-bits/anvill/files/6232772/anvill_issue_150_test_case.json.zip)

Found in commit 1ce66508c92ab146b33716f07ef5a151dce458eb

```
bool InstructionFolderPass::FoldPHINode(
    InstructionFolderPass::InstructionList &output, llvm::Instruction *instr) {
```
```
  // Finally, drop all the instructions we no longer need; the `PHINode`
  // instruction will be deleted later by the caller
  auto function_changed = !inst_replacement_list.empty();
  PerformInstructionReplacements(inst_replacement_list);

  for (auto &incoming_value : incoming_value_list) {
    auto value_as_instr =
        llvm::dyn_cast<llvm::Instruction>(incoming_value.value);

    if (value_as_instr == nullptr) {
      continue;
    }

    if (value_as_instr->use_empty()) {
      value_as_instr->eraseFromParent(); // crash here
    }
  }
```

pgoodman:
![image](https://user-images.githubusercontent.com/43664/113068493-6459fc80-918c-11eb-9d0a-386a1c61ff27.png)

pgoodman:
![image](https://user-images.githubusercontent.com/43664/113068633-a71bd480-918c-11eb-9961-c32fe3f29d64.png)
![image](https://user-images.githubusercontent.com/43664/113068646-b0a53c80-918c-11eb-91b4-178f07f2b83d.png)

pgoodman:
Possible fix:

```c++
  std::unordered_set<llvm::Instruction *> to_remove;
  for (auto &incoming_value : incoming_value_list) {
    auto value_as_instr =
        llvm::dyn_cast<llvm::Instruction>(incoming_value.value);
    if (value_as_instr && value_as_instr->use_empty()) {
      to_remove.insert(value_as_instr);
    }
  }
  for (auto value_as_instr : to_remove) {
    value_as_instr->eraseFromParent();
    function_changed = true;
  }
```

alessandrogario:
Fixed in 0fce7c9a00f464428e58858d6d3fd78c7597bdff

PR #163, opened 2021-04-12
Fix segfault instruction folder pass

kumarak:
The changes fix the segfault in the instruction folder pass and avoid infinite folding of the Phi node due to cast instructions.
Ref: #150

### DIFF
--- a/libraries/anvill_passes/src/InstructionFolderPass.cpp
+++ b/libraries/anvill_passes/src/InstructionFolderPass.cpp
@@ -47,6 +47,7 @@ const std::unordered_map<std::uint32_t, InstructionFolder> kInstructionFolderMap
   { llvm::Instruction::Select, InstructionFolderPass::FoldSelectInstruction },
   { llvm::Instruction::PHI, InstructionFolderPass::FoldPHINode },
 };
+
 // clang-format on
 
 // Case handlers for `SelectInst` instructions
@@ -57,6 +58,7 @@ using SelectInstructionFolder = bool (*)(llvm::Instruction *&output,
 
 // clang-format off
 const std::unordered_map<std::uint32_t, SelectInstructionFolder> kSelectInstructionFolderMap = {
+
   // Binary operators
   { llvm::Instruction::Add, InstructionFolderPass::FoldSelectWithBinaryOp },
   { llvm::Instruction::FAdd, InstructionFolderPass::FoldSelectWithBinaryOp },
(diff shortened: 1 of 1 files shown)


libraries/anvill_passes/src/InstructionFolderPass.cpp
```diff
@@ -133,13 +137,20 @@ const std::unordered_map<std::uint32_t, PHINodeInstructionFolder> kPHINodeFolder
   { llvm::Instruction::SIToFP, InstructionFolderPass::FoldPHINodeWithCastInst },
   { llvm::Instruction::FPTrunc, InstructionFolderPass::FoldPHINodeWithCastInst },
   { llvm::Instruction::FPExt, InstructionFolderPass::FoldPHINodeWithCastInst },
+
+
+  // NOTE: The folding of phi nodes with PtrToInt, IntToPtr, and BitCast is disabled
+  //       in the function since bouncing these into each other can cause folding
+  //       to enter into infinite loop.
+  //
   { llvm::Instruction::PtrToInt, InstructionFolderPass::FoldPHINodeWithCastInst },
```
pgoodman, line 146: Can you disable via commenting our here, rather than checking with an `if` statement below?

libraries/anvill_passes/src/InstructionFolderPass.cpp
```diff
@@ -275,33 +288,55 @@ void InstructionFolderPass::PerformInstructionReplacements(
   }
 }
 
+// If there is cyclic dependency incoming values on phi node; It can't be folded. The
+// utility function checks such cases.
+static inline bool
+isPHINodeFoldable(llvm::Instruction *phi_node,
```
pgoodman, line 294: Naming scheme: `IsPHINode...`

libraries/anvill_passes/src/InstructionFolderPass.cpp
```diff
@@ -523,11 +561,23 @@ bool InstructionFolderPass::FoldPHINodeWithCastInst(
     destination_type = instr->getDestTy();
   }
 
+  // NOTE(akshayk) : A phi node folding with cast instructions like IntToPtr
+  //                 PtrToInt, and BitCast may enter into an infinite loop
+  //                 if the folding instruction bounce between each other. Added
+  //                 checks to avoid such cases.
+  //
+  if ((cast_instr->getOpcode() == llvm::Instruction::IntToPtr) ||
```
pgoodman, line 569: Lets remove this and just depend on not including them in the map / commenting them out in the map.

pgoodman (changes requested):

pgoodman (approved):

Issue #156, opened 2021-04-08
Incorrect control flow redirection table in spec


kumarak:
The `control_flow_redirection` table is getting generated wrongly for the non-PIE binaries. The redirection destination address is null which is incorrect. The `function_thunk.start` address does not seem to be offset and adding `image_base_address` will generate the wrong address to get the external table.

```
    "control_flow_redirections": [
        [
            4198448,
            null
        ],
        [
            4198464,
            null
        ],
        [ 
            4198480,
            null
        ]
    ],```

PR #165, opened 2021-04-13
Exporter: Update elf thunk resolution

alessandrogario:
Fixes issue https://github.com/lifting-bits/anvill/issues/156

### DIFF
--- a/anvill/python/anvill/binja/bnprogram.py
+++ b/anvill/python/anvill/binja/bnprogram.py
@@ -34,7 +34,11 @@ class BNProgram(Program):
         self._path: Final[str] = path
         self._bv: Final[bn.BinaryView] = bv
         self._type_cache: Final[TypeCache] = TypeCache(self._bv)
-        self._init_ctrl_flow_redirections()
+
+        try:
+            self._init_ctrl_flow_redirections()
+        except:
+            print("Failed to initialize control flow redirections")
 
     @property
     def bv(self):
--- a/anvill/python/anvill/binja/bnprogram.py
+++ b/anvill/python/anvill/binja/bnprogram.py
@@ -26,6 +26,7 @@ from anvill.program import *
 from anvill.arch import *
 from anvill.os import *
 from anvill.imageparser import *
+from anvill.util import *
 
 
 class BNProgram(Program):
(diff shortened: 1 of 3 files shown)


anvill/python/anvill/program.py
```diff
@@ -175,6 +175,10 @@ def proto(self):
             proto["functions"].append(func.proto())
 
         for source, target in self._control_flow_redirections.items():
+            if type(source) != int or type(target) != int:
```
pgoodman, line 178: This is the wrong place to fix it; instead, we should not admit "bad data" into the object in the first place, i.e. `add_control_flow_redirection` should be annotated with types `from: int, to: int` and then it should be enforced at the source, and we should work our way back to the source of the issue.

pgoodman (changes requested):

kumarak:
The fix will bypass the `null` value but not resolve the actual problem. The null value creeps in because the relocation offset is getting wrong from the elf image.

alessandrogario:
> The fix will bypass the `null` value but not resolve the actual problem. The null value creeps in because the relocation offset is getting wrong from the elf image.

There are two commits in the PR, the first one solves the address problem

kumarak (approved):

anvill/python/anvill/binja/bnprogram.py
```diff
@@ -34,7 +34,11 @@ def __init__(self, bv: bn.BinaryView, path: str):
         self._path: Final[str] = path
         self._bv: Final[bn.BinaryView] = bv
         self._type_cache: Final[TypeCache] = TypeCache(self._bv)
-        self._init_ctrl_flow_redirections()
+
+        try:
+            self._init_ctrl_flow_redirections()
+        except:
+            print("Failed to initialize control flow redirections")
```
pgoodman, line 41: File an issue that we eventually need a better logging approach.

pgoodman (approved):

anvill/python/anvill/binja/bnprogram.py
alessandrogario: Opened it here: https://github.com/lifting-bits/anvill/issues/166

pgoodman (approved):

PR #169, opened 2021-04-14
Update control flow target support

alessandrogario:


anvill/src/Lifters/FunctionLifter.cpp
```diff
@@ -115,6 +116,31 @@ static void MuteStateEscape(llvm::CallInst *call) {
   call->setArgOperand(remill::kStatePointerArgNum, undef_val);
 }
 
+static llvm::Function *GetAnvillSwitchFunc(llvm::Module &module,
```
pgoodman, line 119: Add a comment here describing the purpose of this function and showing an example of what a call to this function looks like and how it's used.

anvill/src/Lifters/FunctionLifter.cpp
```diff
@@ -115,6 +116,31 @@ static void MuteStateEscape(llvm::CallInst *call) {
...
+                                           bool complete) {
+
+  auto func_name =
```
pgoodman, line 122: `const auto &`, otherwise it's making a copy of one of the `std::string`s.

anvill/src/Lifters/FunctionLifter.cpp
```diff
@@ -115,6 +116,31 @@ static void MuteStateEscape(llvm::CallInst *call) {
...
+      complete ? kAnvillSwitchCompleteFunc : kAnvillSwitchIncompleteFunc;
+
+  if (auto func = module.getFunction(func_name)) {
```
pgoodman, line 125: move the auto assignment out, so that the later `auto func = ` can just be `func = `.

anvill/src/Lifters/FunctionLifter.cpp
```diff
@@ -115,6 +116,31 @@ static void MuteStateEscape(llvm::CallInst *call) {
...
+
+  auto &context = module.getContext();
+  auto return_type = llvm::Type::getInt64Ty(context);
+  const std::vector<llvm::Type *> func_parameters = {
+      llvm::Type::getInt64Ty(context)};
+
+  auto func_type = llvm::FunctionType::get(return_type, func_parameters, true);
+
+  auto func = llvm::Function::Create(
+      func_type, llvm::GlobalValue::ExternalLinkage, func_name.c_str(), module);
```
pgoodman, line 137: I don't think you need to do `.c_str()`. It should be an `llvm::StringRef`, which can be implicitly constructed from a `std::string`.

anvill/src/Lifters/FunctionLifter.cpp
```diff
@@ -260,16 +290,106 @@ void FunctionLifter::VisitDirectJump(const remill::Instruction &inst,
   llvm::BranchInst::Create(GetOrCreateTargetBlock(inst.branch_taken_pc), block);
 }
 
-// Visit an indirect jump control-flow instruction. This may be register- or
-// memory-indirect, e.g. `jmp rax` or `jmp [rax]` on x86. Thus, the target is
-// not know a priori and our default mechanism for handling this is to perform
-// a tail-call to the `__remill_jump` function, whose role is to be a stand-in
-// something that enacts the effect of "transfer to target."
+// This is a list of the possibilities we want to cover:
```
pgoodman, line 293: Update the comment to include part of the old comment. I'd like these comments to provide the kind of end-to-end context, i.e. this thing is related to `jmp rax`. Your current comment is good at describing the mechanics/corner cases.

anvill/src/Lifters/FunctionLifter.cpp
```diff
@@ -260,16 +290,106 @@ void FunctionLifter::VisitDirectJump(const remill::Instruction &inst,
...
+
+  auto add_remill_jump{true};
+  llvm::BasicBlock *current_bb = block;
+
+  if (maybe_target_list.has_value()) {
+    const auto &target_list = maybe_target_list.value();
+
+    if (target_list.destination_list.size() == 1U && target_list.complete) {
+
+      // If the target list is complete and has only one destination, then we
```
pgoodman, line 320: I prefer putting these types of comments before the `if` or the `else if`, i.e. the comment describes the condition and the body if the condition is true. The code formatting script has stuff to properly handle this commenting style.

anvill/src/Lifters/FunctionLifter.cpp
```diff
@@ -260,16 +290,106 @@ void FunctionLifter::VisitDirectJump(const remill::Instruction &inst,
...
+      // can handle it as normal jump
+      add_remill_jump = false;
+
+      auto destination = target_list.destination_list.front();
+      llvm::BranchInst::Create(GetOrCreateTargetBlock(destination), block);
+
+    } else {
+
+      // We have multiple destinations. Handle this with a switch. If the target
```
pgoodman, line 329: Relatd to above, I would move this comment to  be just before the `} else {`.

pgoodman (changes requested):

anvill/src/TypeParser.cpp
```diff
@@ -541,6 +543,11 @@ llvm::Expected<llvm::Type *> ParseType(llvm::LLVMContext &context,
     }
 
     auto type = remill::GetReference(ret);
+    if (type->isFunctionTy()) {
```
pgoodman, line 546: I think the better approach is to always return the type, regardless of if it is a function/sized/unsized type, and instead do the `isSized` check at the usage site of `ParseType`. Right now, that logic is basically having to happen with the `is_function` checks at the usage site, so really, that tells me that the `isSized()` check doesn't belong here anymore.

pgoodman (changes requested):

pgoodman (approved):

PR #171, opened 2021-04-22
Resolve function thunk name

kumarak:
The changes resolve the name of function thunks referenced by the IR.


anvill/python/anvill/arch.py
```diff
@@ -35,6 +35,9 @@ def return_stack_pointer_proto(self, num_bytes_popped):
     def pointer_size(self):
         raise NotImplementedError()
 
+    def stack_offset(self):
```
pgoodman, line 38: Can you import `abc` and then make `Arch` derive from `ABC`, then change all these things throwing `NotImplementedError` to use the `@abstractmethod` decorator, then make the function bodies `...`, e.g.

```python
@abstractmethod
def foo():
    ...
```

The `...` is actually itself there.

pgoodman: Also, add return type hints where clear/obvious.

anvill/python/anvill/binja/bnprogram.py
```diff
@@ -36,10 +35,15 @@ def __init__(self, bv: bn.BinaryView, path: str):
...
         self._type_cache: Final[TypeCache] = TypeCache(self._bv)
 
+        # Create an image parser
+        self._image_parser = create_elf_image_parser(bv.file.filename)
+
         try:
             self._init_func_thunk_ctrl_flow()
         except:
-            DEBUG("Failed to initialize the control flow information for functin thunks")
+            print(
```
pgoodman, line 44: Change back to `DEBUG`?

anvill/python/anvill/binja/bnprogram.py
```diff
@@ -78,6 +82,114 @@ def get_variable_impl(self, address):
 
         return BNVariable(bn_var, arch, address, var_type)
 
+    def _get_function_from_bnvariable(self, address, bn_var):
+        """Get the `Function` instance from the data variable of function
+        pointer type. Raise an `InvalidFunctionException` exception if
+        the data variable is not of function pointer type.
+        """
+
+        def _is_function_pointer(bn_var):
```
pgoodman, line 91: Make this a function at the top level?

anvill/python/anvill/binja/bnprogram.py
```diff
@@ -78,6 +82,114 @@ def get_variable_impl(self, address):
...
+
+            elif bn.TypeClass.NamedTypeReferenceClass == var_type.type_class:
+                # The function paramater could be named alias of a float type.
+                if isinstance(arg_type, FloatingPointType):
+                    reg_name = calling_conv.next_float_arg_reg
+                else:
+                    reg_name = calling_conv.next_int_arg_reg
+
+            elif bn.TypeClass.VoidTypeClass == var_type.type_class:
+                reg_name = "invalid void"
```
pgoodman, line 149: This isn't a valid register name...

anvill/python/anvill/binja/bnprogram.py
```diff
@@ -78,6 +82,114 @@ def get_variable_impl(self, address):
...
+
+            else:
+                reg_name = None
+                raise AnvillException(
+                    "No variable type defined for function parameters"
+                )
+
+            if reg_name:
+                loc = Location()
+                loc.set_register(reg_name.upper())
```
pgoodman, line 159: Instead of `reg_name.upper()`, you should use the `arch.register_name(reg_name)` method; that is why it is there. If yo usee other uses of `.upper()`, then please transition them to `arch.register_name(reg_name)`; using `.upper()` is not always the right approach.

anvill/python/anvill/binja/bnprogram.py
```diff
@@ -78,6 +82,114 @@ def get_variable_impl(self, address):
...
+                loc.set_type(arg_type)
+                param_list.append(loc)
+
+            else:
+                loc = Location()
+                loc.set_memory(self._bv.arch.stack_pointer.upper(), stack_offset)
```
pgoodman, line 165: Ditto to the base reg here.

anvill/python/anvill/binja/bnprogram.py
```diff
@@ -78,6 +82,114 @@ def get_variable_impl(self, address):
...
+                stack_offset = stack_offset + arch.pointer_size()
+
+            index += 1
+
+        ret_list = []
+        retTy = self.type_cache.get(func_tinfo.return_value)
+        if not isinstance(retTy, VoidType):
+            for reg in calling_conv.return_regs:
+                loc = Location()
+                loc.set_register(reg.upper())
```
pgoodman, line 177: Ditto here

anvill/python/anvill/binja/bnprogram.py
```diff
@@ -78,6 +82,114 @@ def get_variable_impl(self, address):
...
+                loc = Location()
+                loc.set_memory(self._bv.arch.stack_pointer.upper(), stack_offset)
+                loc.set_type(arg_type)
+                param_list.append(loc)
+                stack_offset = stack_offset + arch.pointer_size()
+
+            index += 1
+
+        ret_list = []
+        retTy = self.type_cache.get(func_tinfo.return_value)
```
pgoodman, line 173: naming convention is `ret_ty`

anvill/python/anvill/binja/bnprogram.py
pgoodman: I don't quite understand the logic here. If the return type is an integer, then will it appear in `rax`, `rdx`, `rcx`, `st0`, `st1`, etc. for amd64?

anvill/python/anvill/binja/bnprogram.py
```diff
@@ -199,7 +315,11 @@ def _init_func_thunk_ctrl_flow(self):
             # Get the variable defined at the dest address
             func_location = self._bv.get_data_var_at(function_thunk.start)
             if not func_location:
-                print("anvill: No variable defined for {:x}/{}".format(function_thunk.start, function_thunk.name))
+                print(
```
pgoodman, line 318: `DEBUG`

anvill/python/anvill/ida/idavariable.py
```diff
@@ -32,13 +32,11 @@ def visit(self, program, is_definition, add_refs_as_defs):
         seg_ref = [None]
         seg = find_segment_containing_ea(self.address(), seg_ref)
         if seg and is_imported_table_seg(seg):
-            print(
-                "Variable at {:x} is in an import table!".format(self.address())
-            )
+            print("Variable at {:x} is in an import table!".format(self.address()))
```
pgoodman, line 35: `DEBUG`

tools/decompile-json/src/main.cpp
```diff
@@ -1036,7 +1036,16 @@ int main(int argc, char *argv[]) {
       }
     } else if (fdecl) {
       if (auto func = lifter.DeclareEntity(*fdecl)) {
+        auto func_with_name = module.getFunction(name);
```
pgoodman, line 1039: I think the following heuristic could apply:

step 1:
iterate over all named things, and apply the following naming scheme: `_<zero-padded hex address to width 8>_name`. (use `std::setw(8) << std::setfill('0') << std::hex << addr` to the `std::stringstream`.

step 2:
iterate over all named things. If the named thing is an external declaration that is called, set its name to `<name>`.

step 3:
iterate over all named things. if the entity is not a declaration, and if the name isn't already taken, then give the thing the name `<name>`

step 4:
iterate over all named things. If the thing's name isn't `<name>` and if its name starts with `_`, and if nothing else is using the name, then give the name the name `<name>`.

pgoodman (changes requested):

tools/decompile-json/src/main.cpp
kumarak: I updated the symbol naming heuristic along the same line and tested with a few binaries which should work. It goes through the functions and fixes the symbol after the first iteration.

anvill/python/anvill/binja/bnprogram.py
kumarak: It will raise an exception for void-type arguments instead of using invalid register name.

anvill/python/anvill/arch.py
```diff
@@ -13,32 +13,45 @@
...
-    def return_address_proto(self):
-        raise NotImplementedError()
+    @abstractmethod
+    def return_address_proto(self) -> dict:
+        ...
 
-    def return_stack_pointer_proto(self, num_bytes_popped):
-        raise NotImplementedError()
+    @abstractmethod
+    def return_stack_pointer_proto(self, num_bytes_popped) -> dict:
```
pgoodman, line 39: `Dict` is the typename, curiously. Be sure to import it from `typing`.

anvill/python/anvill/arch.py
```diff
@@ -13,32 +13,45 @@
...
 
-    def register_family(self, reg_name):
-        raise NotImplementedError()
+    @abstractmethod
+    def stack_offset(self) -> int:
+        ...
 
-    def register_name(self, reg_name):
+    @abstractmethod
+    def register_family(self, reg_name) -> tuple:
```
pgoodman, line 51: It should be `Tuple` (also needs importing) and maybe something like: `Tuple[blah, ...]` where the `blah` is the type name for the first thing, and the `...` signals that there is a variable number of things in the tuple, and each subsequent thing has the same type as `blah`.

anvill/python/anvill/arch.py
```diff
@@ -333,9 +346,15 @@ def return_stack_pointer_proto(self, num_bytes_popped):
     def pointer_size(self):
```
pgoodman, line 346: `-> int:` ret type

anvill/python/anvill/arch.py
```diff
@@ -333,9 +346,15 @@ def return_stack_pointer_proto(self, num_bytes_popped):
...
         return 8
 
+    def stack_offset(self):
```
pgoodman, line 349: `-> int:` ret type

anvill/python/anvill/arch.py
```diff
@@ -333,9 +346,15 @@ def return_stack_pointer_proto(self, num_bytes_popped):
...
+        return 8
+
     def register_family(self, reg_name):
```
pgoodman, line 352: Return type

anvill/python/anvill/arch.py
```diff
@@ -333,9 +346,15 @@ def return_stack_pointer_proto(self, num_bytes_popped):
...
         return self._REG_FAMILY[reg_name]
 
+    def register_name(self, reg_name):
```
pgoodman, line 355: `-> str:` return type

anvill/python/anvill/arch.py
```diff
@@ -569,9 +588,15 @@ def return_stack_pointer_proto(self, num_bytes_popped):
     def pointer_size(self):
         return 4
 
+    def stack_offset(self):
```
pgoodman, line 591: `-> int:` return type

anvill/python/anvill/arch.py
```diff
@@ -569,9 +588,15 @@ def return_stack_pointer_proto(self, num_bytes_popped):
     def pointer_size(self):
```
pgoodman, line 588: `-> int:` return type

anvill/python/anvill/arch.py
```diff
@@ -569,9 +588,15 @@ def return_stack_pointer_proto(self, num_bytes_popped):
...
         return 4
 
+    def stack_offset(self):
+        return 4
+
     def register_family(self, reg_name):
         return self._REG_FAMILY[reg_name]
 
+    def register_name(self, reg_name):
```
pgoodman, line 597: `-> str:` return type

anvill/python/anvill/arch.py
```diff
@@ -942,37 +967,50 @@ def return_stack_pointer_proto(self, num_bytes_popped):
     def pointer_size(self):
         return 8
 
+    def stack_offset(self):
```
pgoodman, line 970: etc.

anvill/python/anvill/arch.py
```diff
@@ -942,37 +967,50 @@ def return_stack_pointer_proto(self, num_bytes_popped):
...
+        return 0
+
+    def register_family(self, reg_name):
+        if reg_name.startswith("%"):
```
pgoodman, line 974: The `%` logic should go in `register_name`

anvill/python/anvill/binja/bnprogram.py
```diff
@@ -75,9 +86,137 @@ def get_variable_impl(self, address):
...
+            if var_type is None:
+                continue
+
+            arg_type = self.type_cache.get(var_type)
+
+            if (
+                bn.TypeClass.IntegerTypeClass == var_type.type_class
+                or bn.TypeClass.PointerTypeClass == var_type.type_class
+            ):
+                reg_name = calling_conv.next_int_arg_reg
```
pgoodman, line 142: What happens if we run out of integer argument registers? Should be be re-thinking this, perhaps letting our function specs either declare: returns+params, OR declare a single function type, then allow our ABI handling logic on the C++ side to do the work?

anvill/python/anvill/binja/callingconvention.py
```diff
@@ -13,36 +13,41 @@
 # You should have received a copy of the GNU Affero General Public License
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
+import binaryninja as bn
 
-class CallingConvention:
+from anvill.util import *
+
+_FLOAT_ARGS_REGS = {
```
pgoodman, line 20: Should some of these be uppercase?

pgoodman (changes requested):

anvill/python/anvill/binja/callingconvention.py
kumarak: The register's name gets changed to uppercase at a later point before the register family lookup or generating the spec. The binary ninja calling convention also gives the register's name in lower case and it gets handled accordingly at a later point.

anvill/python/anvill/binja/bnprogram.py
kumarak: Here we approximate the ABI while generating spec and follow what the calling convention says. There are instances in binja where the function type does not follow ABI. We discard arguments and force them to follow the ABI. It comes down to how reliable is decompiler in function type identification and may decide to follow that and let the C++ side handle the ABI logic.

anvill/python/anvill/arch.py
```diff
@@ -48,7 +49,7 @@ def stack_offset(self) -> int:
         ...
 
     @abstractmethod
-    def register_family(self, reg_name) -> tuple:
+    def register_family(self, reg_name) -> Tuple[Tuple[str, ...]]:
```
pgoodman, line 52: I think it's `Tuple[Tuple[str, int, int], ...]`?

anvill/python/anvill/arch.py
```diff
@@ -328,32 +329,38 @@ class AMD64Arch(Arch):
...
+    def pointer_size(self) -> int:
         return 8
 
-    def stack_offset(self):
+    def stack_offset(self) -> int:
         return 8
 
-    def register_family(self, reg_name):
-        return self._REG_FAMILY[reg_name]
+    def register_family(self, reg_name) -> Tuple[Tuple[str, ...]]:
```
pgoodman, line 353: Fix ret type

anvill/python/anvill/arch.py
pgoodman: Maybe what you can do is assign something at the top level, e.g. `RegisterBounds = Tuple[Register, int, int]`, then you can do `-> Tuple[RegisterBounds, ...]`

anvill/python/anvill/arch.py
```diff
@@ -570,32 +577,38 @@ class X86Arch(Arch):
         "ST7": (("ST7", 0, 10), ("ST7", 0, 12), ("ST7", 0, 16)),
     }
 
-    def name(self):
+    def name(self) -> str:
```
pgoodman, line 580: Lets update the top level to have `ArchName = str`, and start using that in place of `str` for the architecture names

anvill/python/anvill/arch.py
```diff
@@ -328,32 +329,38 @@ class AMD64Arch(Arch):
...
```
pgoodman, line 332: Lets update the top level to have `ArchName = str`, and start using that in place of `str` for the architecture names

pgoodman (changes requested):

anvill/python/anvill/arch.py
```diff
@@ -1005,127 +1066,125 @@ class AArch32Arch(Arch):
...
-    def pointer_size(self):
+    def pointer_size(self) -> int:
         return 4
 
-    def register_family(self, reg_name):
+    def stack_offset(self) -> int:
+        return 0
+
+    def register_family(self, reg_name) -> Tuple[RegisterBounds, ...]:
         if reg_name.startswith("%"):
```
pgoodman, line 1178: Here we should do: `self._REG_FAMILY[self.register_name(reg_name)]`, and/or have the `reg_name` parameter of `register_family` be a `Register`.

anvill/python/anvill/arch.py
```diff
@@ -1452,31 +1514,34 @@ class Sparc64Arch(Arch):
...
+    def pointer_size(self) -> int:
         return 8
 
-    def register_family(self, reg_name):
+    def stack_offset(self) -> int:
+        return 2227
+
+    def register_family(self, reg_name) -> Tuple[RegisterBounds, ...]:
         if reg_name.startswith("%"):
             return self._REG_FAMILY[reg_name[1:].lower()]
```
pgoodman, line 1540: Ditto here.

pgoodman (changes requested):

anvill/python/anvill/arch.py
kumarak: Thanks for catching this. It's a copy-paste issue from the `register_name`.

kumarak: Thanks for catching this. `register_family` implementation had redundant code. I updated it to use `register_name` directly. I think let the parameter as string type for the user and get `register_name` internally.

anvill/python/anvill/binja/bnprogram.py
```diff
@@ -83,9 +95,58 @@ def get_variable_impl(self, address: int):
...
+
+        arch = self._arch
+        func_tinfo = bn_var.type.element_type
+
+        # TODO(akshayk): The type information may not have calling convention
+        # information. It does not recover the parameters and get the function
+        # in such case. A working solution could be handing the default calling
+        # convention for each architecture. This is in todo list.
+        if func_tinfo.calling_convention is None:
+            return None
```
pgoodman, line 134: Assume calling convention `0`, which we'll default to the os/arch specific one.

anvill/python/anvill/binja/bnprogram.py
```diff
@@ -83,9 +95,58 @@ def get_variable_impl(self, address: int):
...
+
+        calling_conv = CallingConvention(arch, func_tinfo)
+
+        func_type = self.type_cache.get(func_tinfo)
+
+        # Get the start address of function which is assigned to global variable
+        is_64bit = self._bv.arch.address_size == 8
+        binary_reader = bn.BinaryReader(self._bv, bn.Endianness.LittleEndian)
+        binary_reader.seek(address)
+        function_start = binary_reader.read64() if is_64bit else binary_reader.read32()
```
pgoodman, line 144: Why is this not just `address`?

anvill/python/anvill/binja/bnprogram.py
```diff
@@ -121,6 +187,24 @@ def get_function_impl(self, address):
...
+                # register allocation in such cases violate calling convention.
+                #
+                # https://github.com/Vector35/binaryninja-api/issues/2399
+                # https://github.com/Vector35/binaryninja-api/issues/2400
+                #
+                # Identify the storage register and discard the parameter variable
+                # if they does not follow calling convention
+                # e.g: int32_t main(int32_t arg1, void* arg2, int128_t arg3 @ q1, int64_t arg4 @ q2)
+                #
+                storage_reg_name = self._bv.arch.get_reg_name(var.storage)
```
pgoodman, line 201: This is what I was askign about.

anvill/python/anvill/binja/bnprogram.py
```diff
@@ -121,6 +187,24 @@ def get_function_impl(self, address):
...
+                if not (
+                    storage_reg_name in calling_conv.int_arg_reg
+                    or storage_reg_name in calling_conv.float_arg_reg
+                ):
+                    continue
+
                 if (
```
pgoodman, line 208: We need to change a lot of / remove a lot of the below, I assume? That is, it seems like we should only use `storage_reg_name`, and if not, we need to be consistent and make sure that if there was a `storage_reg_name` for some prior parameter, that we don't permit that register name to be used anymore. I.e. the `CallingConvention` should have some kind of used registers tracking, to prevent accidental reuse.

pgoodman: We should throw an exception for that "invalid void" instead of admitting a bad register.

pgoodman: Also, if you find such an example, then report an issue to v35.

anvill/python/anvill/binja/bnprogram.py
```diff
@@ -121,6 +187,24 @@ def get_function_impl(self, address):
...
+                # Identify the storage register and discard the parameter variable
+                # if they does not follow calling convention
+                # e.g: int32_t main(int32_t arg1, void* arg2, int128_t arg3 @ q1, int64_t arg4 @ q2)
+                #
+                storage_reg_name = self._bv.arch.get_reg_name(var.storage)
+                if not (
+                    storage_reg_name in calling_conv.int_arg_reg
+                    or storage_reg_name in calling_conv.float_arg_reg
+                ):
+                    continue
```
pgoodman, line 206: This should be an exception, not a silent ignoring.

pgoodman (changes requested):

anvill/python/anvill/binja/bnprogram.py
kumarak: ref:
https://github.com/Vector35/binaryninja-api/issues/2399
https://github.com/Vector35/binaryninja-api/issues/2400

pgoodman (approved):

Issue #170, opened 2021-04-22
Enable rebasing of binaries on load


artemdinaburg:
Add a new command line option to rebase binaries to a specific base address. The goal of this option is two-fold

1) Load firmware/ROMs/etc that need a specific base address but are specified without one.

2) We use object files (i.e., .o files) that are based at 0x0 for some tests. Rebasing puts them at a more sane address, such that we can use basic heuristics to find undetected references in the final output.

Desired functionality:

Anvill gets a new command line option of `--rebase <hex base address>`. The loaded binary is rebased during load with ida/binja. If needed: add an item to the specification JSON with the original and rebased addresses of the binary.

PR #173, opened 2021-04-23
Exporter: Add base_address option

alessandrogario:
This fixes issue https://github.com/lifting-bits/anvill/issues/170

### DIFF
--- a/anvill/python/anvill/ida/idaprogram.py
+++ b/anvill/python/anvill/ida/idaprogram.py
@@ -14,6 +14,10 @@
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
 
+import argparse
+from typing import Optional
+
+
 import ida_funcs
 import ida_typeinf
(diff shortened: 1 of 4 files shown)


anvill/python/anvill/ida/idaprogram.py
```diff
@@ -48,6 +51,26 @@
 
 class IDAProgram(Program):
     def __init__(self, arch, os):
+        arg_parser = argparse.ArgumentParser()
```
pgoodman, line 54: This should go into `get_program`, and be guarded to execute only once per program. Then the `get_program` function in both architectures should have an optional `base_address` kwarg.

pgoodman (changes requested):

pgoodman (approved):
Approving. @alessandrogario feel free to merge if this doesn't affect your ELF parser, or if your ELF parser has already been adapted.

PR #176, opened 2021-04-27
Disable functions with thumb2 instructions in spec

kumarak:
Disable filling memory bytes for the function having `thumb2` instruction sets in spec.


pgoodman (approved):
Can you file an issue that references this PR, where in the issue, you say something like "add Thumb2 support" and "when added, remove the code from PR #176".

PR #178, opened 2021-04-28
Fix Type Parser

kumarak:
The changes fix type parser for Pointer and Array type. A pointer type can be of void or unsize type. Also, the spec can have an array of size 0.


anvill/src/TypeSpecification.cpp
```diff
@@ -431,11 +431,18 @@ TypeSpecification::ParseType(llvm::LLVMContext &llvm_context,
...
-        }
+        // An array can be of size `0` used in the variable length object. The specs for
+        // them will have size zero. e.g:
+        //     struct dir {
+        //        int32_t fd;           *=0{ii[bx0]}
+        //        int32_t errcode;
+        //        char data[0x0];
+        //     };
+        //
+        // Don't throw error if the number of elements is zero.
```
pgoodman, line 442: Are zero-sized arrays valid in LLVM bitcode? How do these come about?

artemdinaburg: Yes they are valid, I just tested with clang. How do they come about from Anvill 🤷 ? but they are valid.

Example:

```c
#include <stdio.h>

int foo[0];

int main(void) {
  printf("Foo is: %p\n", foo);
  return 0;
}
```

Will result with bitcode that has:
```
@foo = common global [0 x i32] zeroinitializer, align 4
```

kumarak: A type spec with zero-sized array (`*=0{iilllli[bx0]}`) generates it as:

 `%anvill.struct.0.239 = type { i32, i32, i64, i64, i64, i64, i32, [0 x i8] }`

pgoodman: TIL!

PR #193, opened 2021-05-02
Add logger

kumarak:
Add logging framework to python script and enable verbose mode

### DIFF
--- a/anvill/python/anvill/__main__.py
+++ b/anvill/python/anvill/__main__.py
@@ -85,7 +85,7 @@ def main():
             )
 
         except:
-            ERROR("The specified address it not valid: '{}'".format(args.base_address))
+            ERROR(f"The specified address it not valid: '{hex(args.base_address)}'")
             return 1
 
     p = get_program(args.bin_in, maybe_base_address)
--- a/anvill/python/anvill/__main__.py
+++ b/anvill/python/anvill/__main__.py
@@ -79,9 +79,7 @@ def main():
         try:
             maybe_base_address = int(args.base_address, 16)
             DEBUG(
-                "Binary Ninja will attempt to load the image at virtual address 0x{:x}".format(
-                    maybe_base_address
-                )
+                f"Binary Ninja will attempt to load the image at virtual address {hex(maybe_base_address)}"
             )
 
         except:
--- a/anvill/python/anvill/__main__.py
+++ b/anvill/python/anvill/__main__.py
@@ -23,7 +23,7 @@ import platform
 
 import binaryninja as bn
 
-from .util import INIT_DEBUG_FILE
+from .util import config_logger
 from .binja import get_program
 
 
@@ -51,9 +51,16 @@ def main():
 
     arg_parser.add_argument(
         "--log_file",
-        type=argparse.FileType("w"),
-        default=os.devnull,
+        type=str,
+        default=None,
         help="Log to a specific file.",
+    )
+
+    arg_parser.add_argument(
+        "--verbose",
+        action="store_true",
+        default=False,
+        help="Enable debug log for the module",
     )
 
     arg_parser.add_argument(
@@ -64,21 +71,21 @@ def main():
 
     args = arg_parser.parse_args()
 
-    if args.log_file != os.devnull:
-        INIT_DEBUG_FILE(args.log_file)
+    # Configure logger
+    config_logger(args.log_file, args.verbose)
 
     maybe_base_address: Optional[int] = None
     if args.base_address is not None:
         try:
             maybe_base_address = int(args.base_address, 16)
-            print(
+            DEBUG(
                 "Binary Ninja will attempt to load the image at virtual address 0x{:x}".format(
                     maybe_base_address
                 )
             )
 
         except:
-            print("The specified address it not valid: '{}'".format(args.base_address))
+            ERROR("The specified address it not valid: '{}'".format(args.base_address))
             return 1
 
     p = get_program(args.bin_in, maybe_base_address)
--- a/anvill/python/anvill/util.py
+++ b/anvill/python/anvill/util.py
@@ -14,26 +14,67 @@
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
 
-_DEBUG_FILE = None
-_DEBUG_PREFIX = ""
+import os
+import sys
+import logging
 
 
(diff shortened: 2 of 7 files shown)


pgoodman (approved):

anvill/python/anvill/util.py
```diff
@@ -14,26 +14,67 @@
...
+logger.setLevel(logging.WARNING)
 
-def DEBUG_PUSH():
-    global _DEBUG_PREFIX
-    _DEBUG_PREFIX += "  "
+log_format = logging.Formatter("%(asctime)s [%(levelname)s]: %(message)s")
 
+try:
+    # Setup console stream handler
+    stream_handler = logging.StreamHandler(sys.stdout)
```
artemdinaburg, line 33: Random nitpick: shouldn't this be `stderr`?

anvill/python/anvill/util.py
```diff
@@ -14,26 +14,67 @@
...
+import os
+import sys
+import logging
 
 
-def INIT_DEBUG_FILE(file):
-    global _DEBUG_FILE
-    _DEBUG_FILE = file
+# Create a logger
+logger = logging.getLogger(__name__)
```
artemdinaburg, line 23: Does use of `__name__` here always make it so that the logger has the name `anvill.util`? I am not 100% sure, but I don't think name propagates from higher-level modules.

anvill/python/anvill/util.py
```diff
@@ -14,26 +14,67 @@
...
 
+# default log level is set to warning; It can be set
+# to debug by enabling verbose mode
+logger.setLevel(logging.WARNING)
 
-def DEBUG_PUSH():
-    global _DEBUG_PREFIX
-    _DEBUG_PREFIX += "  "
+log_format = logging.Formatter("%(asctime)s [%(levelname)s]: %(message)s")
```
artemdinaburg, line 29: Can we maybe add filename/lineno, kind of like:
```"%(asctime)s [%(levelname)s] [%(filename)s:%(lineno)d] %(message)s"```

File: anvill/python/anvill/util.py [hunk 4]
```diff
@@ -14,26 +14,67 @@
...
+        except Exception as e:
+            logger.warning("Failed to set up log file: {}".format(e))
+
+    # enable verbose mode if it is set
+    if verbose:
+        logger.setLevel(logging.DEBUG)
+
+
+# debug log interface
+DEBUG = logger.debug
```
artemdinaburg, line 68: +1 for these handy macro-like features :)

anvill/python/anvill/util.py
```diff
@@ -14,26 +14,67 @@
...
-    global _DEBUG_FILE
-    if _DEBUG_FILE:
-        _DEBUG_FILE.write("{}{}\n".format(_DEBUG_PREFIX, str(s)))
+def config_logger(logfile, verbose=False):
+    """Set the logger file handler and set the log
+    level to verbose if required
+    """
+
+    if logfile is not None:
+        if not logfile.endswith(".log"):
```
artemdinaburg, line 48: Can this be removed? I am particularly concerned about setting logger to a device file of some kind (e.g. `/dev/stderr`, `/dev/stdout` or similar, which this would break)

anvill/python/anvill/util.py
artemdinaburg: So typically you want the logger name to be a full module name based on the calling module, like "anvill.binja.bnprogram", "anvill.type", etc. so that you can filter logs by module name, and like set a logging config to only log certain modules, or only make some log parts verbose and not all. 

If __name__ is indeed always anvill.util, would it be possible to make a function in here, say, something like:

```python
def createLogger(logger_name):
  logger = logging.getLogger(logger_name)
  #... do setup/config here as needed ...
  return logger.info, logger.warning, logger.error, logger.fatal

#Then at the start of each module, you'd have

from .util import createLogger

INFO, WARN, DEBUG, FATAL = createLogger(__name__)

```

anvill/python/anvill/__main__.py
```diff
@@ -51,11 +51,18 @@ def main():
...
     arg_parser.add_argument(
         "--log_file",
-        type=argparse.FileType("w"),
-        default=os.devnull,
+        type=str,
+        default=None,
         help="Log to a specific file.",
     )
 
+    arg_parser.add_argument(
```
artemdinaburg, line 59: Thank you for adding this!

anvill/python/anvill/__main__.py
```diff
@@ -64,21 +71,21 @@ def main():
...
-            print(
+            DEBUG(
                 "Binary Ninja will attempt to load the image at virtual address 0x{:x}".format(
                     maybe_base_address
                 )
             )
 
         except:
-            print("The specified address it not valid: '{}'".format(args.base_address))
+            ERROR("The specified address it not valid: '{}'".format(args.base_address))
```
artemdinaburg, line 88: So I notice you use .format a few times, just as an FYI, python3.6 and newer has f-strings which are much easier to read. For example, you can just do:

```
ERROR(f"The specified address it not valid: '{args.base_address}'")
```

artemdinaburg (changes requested):
Thanks for taking the time to implement some sane logging! This will be much needed. I have some questions/comments in the review, please take a look when you have time.

anvill/python/anvill/util.py
kumarak: @artemdinaburg, Thanks for the review. You are correct. The logger name here is `anvil.util`. It creates a single logger for all the modules. 

If you would like to have a separate logger for each module, it's a good idea to go ahead with the logger configuration file. I will have it in place and have the `root` logger as default.

kumarak: I update the design and hopefully, it should work as you described. The logger can be configured with the `logging.ini` config file. If the file is missing, it falls back to the basic configuration.

There is an interface to create a module-level logger with the default configuration and It can be configured with `logging.ini` file. Please let me know your thoughts.

artemdinaburg (approved):
Thanks Akshay, this looks good, I will have to try it out and see how it logs.

In the future, I want to migrate more to module-level create_logger() calls versus using a global debug/info/warn/etc, but this works with the code we have now and provides a way to get to module level calls.

Thanks again for taking the time to add much needed logging sanity.

PR #194, opened 2021-05-04
Carson/fix pointer lifter

carsonharmon:
This PR does a few things. 

1. Removes some uses of ReplaceAllUses. These were not working anyway as we don't change unless its semantics preserving. 
2. Adds a previously unhandled case to BinaryOperator 
3. Changes the insert position of phi nodes to produce valid bitcode
4. Fixes an infinite loop where PhiNodes were creating inttoptr's repeatedly. 
5. Adds a debug log statement when a change thats not semantic preserving is attempted (it will always fail)
6. Adds challenge 2 as a testcase to confirm that valid bitcode is always produced. 
7.  Adds .ll files generated from Anvill roundtrip with pointer lifter off to be used as tests. This tries to find semantic breaking errors or invalid bitcode produced by the pointer lifter for these tests
8. Fixes bugs when operating on non add/sub binary operators 

Note, there is a regression in test 1 but thats because the const handling code in visitLoad was commented out before. Another pass might make this work in practice.


libraries/anvill_passes/src/BrightenPointerOperations.cpp
```diff
@@ -294,40 +300,43 @@ PointerLifter::visitBitCastInst(llvm::BitCastInst &inst) {
...
 }
 
 // Use this to flatten a gep
 // Replace all uses of the original gep with the return value (final flattened)
 // Remove spurious instructions with dead code removal.
-llvm::Value *
-PointerLifter::flattenGEP(llvm::GetElementPtrInst *gep) {
+llvm::Value *PointerLifter::flattenGEP(llvm::GetElementPtrInst *gep) {
+
   // FIXME (Carson), delete once we can fix the TODO with getting indexed types
```
pgoodman, line 324: Can you file an issue with more about what this FIXME relates to? If so, update, the comment to reference the issue number.

PR #197, opened 2021-05-12
Update and test docker image in CI

ekilmer:
- [x] Add symlinks for the unversioned tools
- [x] Pull latest version of Remill and install that way (instead of from source)


ekilmer:
@alessandrogario Do you know of an easy way to prevent the CI run killer from killing the Docker workflow added in this PR?

alessandrogario:
> @alessandrogario Do you know of an easy way to prevent the CI run killer from killing the Docker workflow added in this PR?

The script should attempt to terminate any other workflow running for the same PR; the easiest way is to embed the new Docker job inside the existing workflow file (which I think would provide faster cancellation time of all the runs if one of the job fails).

Dockerfile
```diff
@@ -1,80 +1,118 @@
...
+# Needed for sourcing venv
+SHELL ["/bin/bash", "-c"]
+
+COPY . ./
+
+RUN source ${VIRTUAL_ENV}/bin/activate && \
+    cmake -G Ninja -B build -S . \
+        -DANVILL_ENABLE_INSTALL_TARGET=true \
+        -Dremill_DIR:PATH=${LIBRARIES}/lib/cmake/remill \
+        -DANVILL_INSTALL_PYTHON3_LIBS=ON \
```
alessandrogario, line 61: This is kind of deprecated, as it's not a good practice to change system state during configure or build time. The `ANVILL_ENABLE_INSTALL_TARGET` setting can be used to install to a $DESTDIR, which can be later used to create a package.

Ideally we want to install both the native code and the Python module in the same place (vs using a venv for Python and /usr for the rest). If we want to still support installing inside a venv maybe we can create a .sh installer that we ship in the package

.github/workflows/docker.yml
```diff
@@ -0,0 +1,44 @@
...
+    steps:
+    - uses: actions/checkout@v2
+      with:
+        submodules: true
+    - name: Build LLVM ${{ matrix.llvm }} on ${{ matrix.ubuntu }}
+      run: |
+        docker build . -t docker.pkg.github.com/lifting-bits/anvill/anvill-llvm${{ matrix.llvm }}-ubuntu${{ matrix.ubuntu }}-amd64:latest -f Dockerfile --target=dist --build-arg UBUNTU_VERSION=${{ matrix.ubuntu }} --build-arg ARCH=amd64 --build-arg LLVM_VERSION=${{ matrix.llvm }}
+    - name: Test Docker image
+      run: |
+        docker run --rm -v $(pwd):/workspace -w /workspace docker.pkg.github.com/lifting-bits/anvill/anvill-llvm${{ matrix.llvm }}-ubuntu${{ matrix.ubuntu }}-amd64:latest -spec /workspace/tools/decompile-json/tests/specs/ret0.json -bc_out /tmp/ret0.bc -ir_out /tmp/ret0.ir
```
alessandrogario, line 31: These tests could be run from within the container when building anvill, instead of repeating them here

Dockerfile
```diff
@@ -1,80 +1,118 @@
...
+        -DCMAKE_INSTALL_PREFIX:PATH="${LIBRARIES}" \
+        -DCMAKE_VERBOSE_MAKEFILE=True \
+        -DVCPKG_ROOT=/dependencies/vcpkg_ubuntu-${UBUNTU_VERSION}_llvm-${LLVM_VERSION}_amd64 \
+        && \
+    cmake --build build --target install
```
alessandrogario, line 66: Once `ANVILL_INSTALL_PYTHON3_LIBS` is removed, both the native library and the Python module can be installed with:

```
cd build

mkdir destdir
export DESTDIR="$(pwd)/destdir"
anvill_version=$(git describe --always)

cmake --build . --target install
cpack -D ANVILL_DATA_PATH="${DESTDIR}" -R ${anvill_version} --config ../packaging/main.cmake

dpkg -i *.deb
```

.github/workflows/docker.yml
artemdinaburg: I think this is to sanity check that the built image works, entry point and all, and is not meant as a replacement for anvill tests.

Dockerfile
artemdinaburg: I think we may also be able to simplify this with `cpack -G DEB`; will take a look

Dockerfile
artemdinaburg: I would very much like to support the venv case for our Python code, it was  much easier to integrate with Binja that way. We should be able to get rid of INSTALL_PYTHON3_LIBS with the fixes you merged last night.

Dockerfile
artemdinaburg: We also specifically want to install into a venv here, so the regular package installation process wont do it.

artemdinaburg: Removed `ANVILL_INSTALL_PYTHON3_LIBS`

anvill/python/install.cmake
```diff
@@ -8,14 +8,20 @@
...
+
+# Only redirect installation if DESTDIR is specified, otherwise just do a normal "setup.py install"
 if(DEFINED ENV{DESTDIR})
-  set(root_parameter "--root=$ENV{DESTDIR}")
+  set(package_parameters 
+    "--single-version-externally-managed"
+    "--prefix=${CMAKE_INSTALL_PREFIX}"
+    "--root=$ENV{DESTDIR}"
+    "--install-purelib=${pure_lib_dest}"
+    )
```
artemdinaburg, line 20: @alessandrogario can you sanity check this for me? I am afraid I may be doing the wrong thing. What I am after is a normal "setup.py install" when DESTDIR does not need to redirect installation.

.github/workflows/docker.yml
ekilmer: Yeah, this is to sanity check that the multi-stage Docker image actually has everything needed to run our installed artifacts....

Thinking more. It would be a nice test to install the packaged deb file that we build, and then do these sanity checks. Right now we just copy the `/opt/trailofbits` directory 🤔

Dockerfile
ekilmer: Hah. Posted this comment before getting to this one https://github.com/lifting-bits/anvill/pull/197#discussion_r632073716

Yes, I like your idea @alessandrogario

artemdinaburg (approved):

PR #205, opened 2021-05-21
Gate builds by testrun against AnghaBench 1K

artemdinaburg:
* Gate builds by the 1K AnghaBench test
* Currently Only tests Anvill-Python, WIP to add decompile-json support

### DIFF
--- a/.github/workflows/anghabench-after-build.yml
+++ b/.github/workflows/anghabench-after-build.yml
@@ -4,7 +4,7 @@ on:
   # Only run when normal CI steps complete
   # Otherwise we'd just fail to build this anvill branch
   workflow_run:
-    workflows: ["Build"]
+    workflows: ["Build On PR"]
     types: [completed]
 
 jobs:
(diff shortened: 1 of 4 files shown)


.github/workflows/anghabench-after-build.yml
```diff
@@ -4,7 +4,7 @@ on:
   # Only run when normal CI steps complete
   # Otherwise we'd just fail to build this anvill branch
   workflow_run:
-    workflows: ["Build"]
+    workflows: ["Build On PR"]
```
ekilmer, line 7: I think this should be reverted since the main workflow name didn't change. Am I missing something?

ekilmer (changes requested):
GitHub Actions runners are really slow running the anghabench test, 36 minutes. Maybe we should look into enabling automatic merges after successful CI pipelines https://docs.github.com/en/github/collaborating-with-issues-and-pull-requests/incorporating-changes-from-a-pull-request/automatically-merging-a-pull-request

Screen Shot 2021-05-24 at 12 29 04 PM

.github/workflows/anghabench-after-build.yml
artemdinaburg: Yes this looks like an accident

ekilmer (approved):

PR #208, opened 2021-05-27
Carson/more pointer bugs

carsonharmon:
There was some commented out code which prevented ptrtoint instructions from being brightened in the case of some other failure. This ended up blocking the whole lifting process. 

The testcases that currently exist check to make sure that everything passed through the pointer lifter produces valid bitcode, not necessarily brightened bitcode. 

This PR also adds a small new brightening pattern for ptrtoint/add without needing any inttoptr's. The test for that features is written inside multiple_bitcast.ll existing case. 

Also adds a testcase for rx-brake routine. 

I'm thinking of ways to do that right now, but thought I would make a PR so anyone could view progress/comment

### DIFF
--- a/libraries/anvill_passes/src/BrightenPointerOperations.cpp
+++ b/libraries/anvill_passes/src/BrightenPointerOperations.cpp
@@ -524,18 +524,22 @@ PointerLifter::visitPtrToIntInst(llvm::PtrToIntInst &inst) {
   // was once again casted to be a pointer, maybe of a different type.
   // Try and propagate if possible
 
-  if (auto ptr_inst = llvm::dyn_cast<llvm::Instruction>(&inst)) {
+  if (auto ptr_inst = llvm::dyn_cast<llvm::Instruction>(inst.getOperand(0))) {
     auto [new_ptr, worked] = visitInferInst(ptr_inst, inferred_type);
     if (worked) {
       return {new_ptr, worked};
     }
   }
+  if (auto constant_ptr =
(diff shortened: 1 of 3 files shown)


libraries/anvill_passes/src/BrightenPointerOperations.cpp
```diff
@@ -524,18 +524,22 @@ PointerLifter::visitPtrToIntInst(llvm::PtrToIntInst &inst) {
...
   // Try and propagate if possible
 
-  if (auto ptr_inst = llvm::dyn_cast<llvm::Instruction>(&inst)) {
+  if (auto ptr_inst = llvm::dyn_cast<llvm::Instruction>(inst.getOperand(0))) {
     auto [new_ptr, worked] = visitInferInst(ptr_inst, inferred_type);
     if (worked) {
       return {new_ptr, worked};
     }
   }
+  if (auto constant_ptr =
```
pgoodman, line 533: Use `else if`. Also need to handle `llvm::Argument`?

pgoodman (changes requested):

libraries/anvill_passes/src/BrightenPointerOperations.cpp
carsonharmon: 👍 Now I just check the operand for matching types if its not another instruction to visit, if not it hits the fall through and forces a cast.

pgoodman (approved):

Issue #222, opened 2021-06-11
Size of register ST0 of return value in function declaration at 0 is too small (8 bytes) for value of size 10 bytes


artemdinaburg:
Getting the following failure on x86 anghabench:

```
E0611 09:40:41.015743 491149 main.cpp:461] Size of register ST0 of return value in function declaration at 0 is too small (8 bytes) for value of size 10 bytes
```

Attaching source JSON

[output_st0_size.tar.gz](https://github.com/lifting-bits/anvill/files/6640118/output_st0_size.tar.gz)

artemdinaburg:
Referencing relevant remill issue: https://github.com/lifting-bits/remill/issues/527

PR #227, opened 2021-06-17
Fix type mismatch for ST register

kumarak:
Lift the type spec of floating-point (10 & 12 bytes) as double to avoid type mismatch error for ST registers. 
Ref: #222

### DIFF
--- a/anvill/src/TypeSpecification.cpp
+++ b/anvill/src/TypeSpecification.cpp
@@ -722,7 +722,15 @@ TypeSpecification::ParseType(llvm::LLVMContext &llvm_context,
         i += 1;
         return llvm::Type::getDoubleTy(llvm_context);
       case 'D':  // long double.
+        // The type spec for floating point with size 10 and 12 bytes can be
+        // created as fp80 and double type. If the type is mapped to the
+        // registers create a double type else create fp80 type. This is to
+        // required to avoid type mismatch with remill registers.
         i += 1;
+        if (i == 1) {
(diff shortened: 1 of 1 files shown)


pgoodman (approved):

anvill/src/TypeSpecification.cpp
```diff
@@ -723,7 +723,10 @@ TypeSpecification::ParseType(llvm::LLVMContext &llvm_context,
         return llvm::Type::getDoubleTy(llvm_context);
       case 'D':  // long double.
         i += 1;
-        return llvm::Type::getX86_FP80Ty(llvm_context);
+        // The type spec for floating point with size 10 and
+        // 12 bytes. Create a double type to avoid type mismatch
+        // for ST registers in remill
+        return llvm::Type::getDoubleTy(llvm_context);
```
pgoodman, line 729: You need to check that this isn't embedded inside of a pointer or structure. If we have a pointer to a `long double`, or a structure containing a `long double` then I think we should leave it as is.

pgoodman (changes requested):

anvill/src/TypeSpecification.cpp
pgoodman: Status on this @kumarak?

kumarak: It will check the spec index during parsing to verify if the spec is generated from the primitive types or is a part of the pointer/struct.

PR #228, opened 2021-06-18
Anvill xor converter pass

artemdinaburg:
Convert some specific instances of the form:

```
%c = icmp pred, op1, op2
%x = xor i1 %c , true
```
into:
```
%c = icmp !pred, op1, op2
%x = %c
```

And update uses of BranchInst and SelectInst. Bail out if uses include things that are not branches or selects.

This comes up often enough in lifted bitcode to do this transform.

### DIFF
--- a/libraries/anvill_passes/src/ConvertXorToCmp.cpp
+++ b/libraries/anvill_passes/src/ConvertXorToCmp.cpp
@@ -89,11 +89,9 @@ bool ConvertXorToCmp::runOnFunction(llvm::Function &func) {
         if (lhs_c && lhs_c->getType()->getBitWidth() == 1 &&
             lhs_c->isAllOnesValue() && rhs_cmp) {
           xors.emplace_back(binop);
-        }
-
-        // right side is a constant int 1 (true), left side is a cmp
-        if (rhs_c && rhs_c->getType()->getBitWidth() == 1 &&
+        } else if (rhs_c && rhs_c->getType()->getBitWidth() == 1 &&
             rhs_c->isAllOnesValue() && lhs_cmp) {
+          // right side is a constant int 1 (true), left side is a cmp
           xors.emplace_back(binop);
         }
       }
--- a/libraries/anvill_passes/src/ConvertXorToCmp.cpp
+++ b/libraries/anvill_passes/src/ConvertXorToCmp.cpp
@@ -24,6 +24,7 @@
 #include <llvm/IR/Module.h>
 #include <llvm/Pass.h>
 
+#include <tuple>
 #include <vector>
 
 namespace anvill {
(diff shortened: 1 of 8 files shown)


libraries/anvill_passes/src/ConvertXorToCmp.cpp
```diff
@@ -0,0 +1,198 @@
...
+  std::vector<llvm::BinaryOperator *> xors;
+
+  for (auto &inst : llvm::instructions(func)) {
+
+    // check for binary op
+    if (auto binop = llvm::dyn_cast<llvm::BinaryOperator>(&inst)) {
+
+      // binary op is a xor
+      if (binop->getOpcode() == llvm::Instruction::Xor) {
+
```
pgoodman, line 79: Call `getComparisonOperand` here, otherwise your later `DLOG(ERROR)` isn't really an error?

libraries/anvill_passes/src/ConvertXorToCmp.cpp
```diff
@@ -0,0 +1,198 @@
...
+  auto changed = false;
+  int replaced_items = 0;
+
+  for (auto xori : xors) {
+
+    // find predicate from xor's operands
+    auto cmp = getComparisonOperand(xori);
+    if (!cmp) {
+      DLOG(ERROR)
```
pgoodman, line 109: This isn't actually an error, so it should be removed.

libraries/anvill_passes/src/ConvertXorToCmp.cpp
```diff
@@ -0,0 +1,198 @@
...
+          << "Error: Xor should have a comparison operand, but we can't find it again";
+      continue;
+    }
+
+    std::vector<llvm::BranchInst *> brs_to_invert;
```
pgoodman, line 114: Move these vectors outside of your `for` loop, and clear them inside so that you can re-use the backing storage across iterations of `xori`.

libraries/anvill_passes/src/ConvertXorToCmp.cpp
```diff
@@ -0,0 +1,198 @@
...
+    std::vector<llvm::SelectInst *> selects_to_invert;
+    bool invertible_xor = true;
+
+    //
+    for (auto &U : cmp->uses()) {
+      llvm::Instruction *inst = llvm::dyn_cast<llvm::Instruction>(U.getUser());
+
+      // use is not the existing xor
+      if (inst == xori) {
```
pgoodman, line 123: Double check `!inst`, just in case.

libraries/anvill_passes/src/ConvertXorToCmp.cpp
```diff
@@ -0,0 +1,198 @@
...
+      DLOG(ERROR)
+          << "Error: Xor should have a comparison operand, but we can't find it again";
+      continue;
+    }
+
+    std::vector<llvm::BranchInst *> brs_to_invert;
+    std::vector<llvm::SelectInst *> selects_to_invert;
+    bool invertible_xor = true;
+
+    //
```
pgoodman, line 118: Fill this comment in with an example of the pattern you're trying to find.

libraries/anvill_passes/src/ConvertXorToCmp.cpp
```diff
@@ -0,0 +1,198 @@
...
+
+      llvm::SelectInst *si = llvm::dyn_cast<llvm::SelectInst>(inst);
+      if (si) {
+        selects_to_invert.emplace_back(si);
+        continue;
+      }
+
+      invertible_xor = false;
+      LOG(INFO) << "ConvertXorToCmp: found a non-invertible xor!\n";
+      inst->dump();
```
pgoodman, line 141: Remove.

libraries/anvill_passes/src/ConvertXorToCmp.cpp
```diff
@@ -0,0 +1,198 @@
...
+      }
+
+      llvm::BranchInst *br = llvm::dyn_cast<llvm::BranchInst>(inst);
+      if (br) {
+        brs_to_invert.emplace_back(br);
+        continue;
+      }
+
+      llvm::SelectInst *si = llvm::dyn_cast<llvm::SelectInst>(inst);
+      if (si) {
```
pgoodman, line 134: Here you should be more strict. That is, you should be checking if the use by the `select` is in fact the condition operand, and not a selected value.

pgoodman (changes requested):

libraries/anvill_passes/src/ConvertXorToCmp.cpp
artemdinaburg: So I purposely do not call getComparisonOperand here since I also need the non-comparison operand. I think a more reasonable solution may be to modify getComparisonOperand() to return both operands, since the logic is nearly identical

libraries/anvill_passes/src/ConvertXorToCmp.cpp
artemdinaburg: 👌

libraries/anvill_passes/src/ConvertXorToCmp.cpp
artemdinaburg: reworked getComparisonOperand => getComparisonOperands

pgoodman (approved):

PR #230, opened 2021-06-24
This is an initial implementation of provenance-tracking metadata, al…

pgoodman:
…ong with an uninterpreted register-value tracking function. The metadata unfortunately doesn't reach the final destination module due to the implementation of instruction cloning in remill.


kumarak (approved):
LGTM

PR #241, opened 2021-07-07
Guard against jump target detection not returning a valid jump target

artemdinaburg:
Sometime the jump target anvill is asking about is... NoneType. Do some sanity checking against this edge case.


anvill/python/anvill/binja/table.py
```diff
@@ -140,12 +140,16 @@ def _find_jumps_near(bv, addr):
...
-            if len(info.branches) != 0:
-                candidates.append(jump_addr)
+            if info:
+                if len(info.branches) != 0:
+                    candidates.append(jump_addr)
+                    break
+
+                jump_addr += info.length
+            else:
+                DEBUG(f"Could not get instruction at jump destination: {jump_addr:x}")
```
pgoodman, line 150: Do we want a `break` here or a `continue`?

PR #251, opened 2021-08-12
Propagate instruction metadata in passes

tetsuo-cpp:
@pgoodman 
I wanted to check in and make sure that I have the right idea here. I looked through Anvill's passes for spots where we replace instructions and copied the metadata across.

The remaining passes seem to either be removing instructions entirely (so no need to propagate anything) or the new instructions don't seem to map to an existing one (`RecoverStackFrameInformation` and `SplitStackFrameAtReturnAddress`). So I've left those untouched.

Also, is there a way to verify this patch? From what I can see, the unit tests just run the pass over some LLVM IR and check that it succeeds.

### DIFF
--- a/libraries/anvill_passes/src/InstructionFolderPass.cpp
+++ b/libraries/anvill_passes/src/InstructionFolderPass.cpp
@@ -263,7 +263,7 @@ bool InstructionFolderPass::FoldSelectInstruction(
 
     // Search for a function that knows how to handle this case
     auto instr_folder_it =
-        kSelectInstructionFolderMap.find(repl.original_instr->getOpcode());
+      kSelectInstructionFolderMap.find(repl.original_instr->getOpcode());
 
     if (instr_folder_it == kSelectInstructionFolderMap.end()) {
       continue;
@@ -463,10 +463,13 @@ bool InstructionFolderPass::FoldSelectWithBinaryOp(
       static_cast<llvm::Instruction::BinaryOps>(binary_op_instr->getOpcode());
 
   auto new_true_value = builder.CreateBinOp(opcode, true_value, operand);
+  CopyMetadataTo(binary_op_instr, new_true_value);
   auto new_false_value = builder.CreateBinOp(opcode, false_value, operand);
+  CopyMetadataTo(binary_op_instr, new_false_value);
 
-  auto replacement = builder.CreateSelect(condition, new_true_value,
-                                          new_false_value, "", select_instr);
+  auto replacement =
+      builder.CreateSelect(condition, new_true_value, new_false_value);
+  CopyMetadataTo(select_instr, replacement);
 
   output = llvm::dyn_cast<llvm::Instruction>(replacement);
   return true;
@@ -506,6 +509,7 @@ bool InstructionFolderPass::FoldPHINodeWithBinaryOp(
     IncomingValue new_incoming_value;
     new_incoming_value.value =
         builder.CreateBinOp(opcode, incoming_value.value, operand);
+    CopyMetadataTo(binary_op_instr, new_incoming_value.value);
 
     new_incoming_value.basic_block = incoming_value.basic_block;
     new_incoming_values.push_back(std::move(new_incoming_value));
@@ -548,12 +552,15 @@ bool InstructionFolderPass::FoldSelectWithCastInst(
 
   auto new_true_value =
       builder.CreateCast(cast_opcode, true_value, destination_type);
+  CopyMetadataTo(cast_instr, new_true_value);
 
   auto new_false_value =
       builder.CreateCast(cast_opcode, false_value, destination_type);
+  CopyMetadataTo(cast_instr, new_false_value);
 
-  auto replacement = builder.CreateSelect(condition, new_true_value,
-                                          new_false_value, "", select_instr);
+  auto replacement =
+      builder.CreateSelect(condition, new_true_value, new_false_value);
+  CopyMetadataTo(select_instr, replacement);
 
   output = llvm::dyn_cast<llvm::Instruction>(replacement);
   return true;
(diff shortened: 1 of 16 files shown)


libraries/anvill_passes/src/LowerRemillMemoryAccessIntrinsics.cpp
```diff
@@ -66,7 +68,9 @@ static void ReplaceMemWriteOp(llvm::CallBase *call_inst, llvm::Type *val_type) {
     val = ir.CreateFPExt(val, val_type);
   }
 
-  (void) ir.CreateStore(val, ptr);
+  val = ir.CreateStore(val, ptr);
```
tetsuo-cpp, line 71: My understanding is that the calls to `ir` above this line are also generating instructions. Would you expect those to also be tagged with this same metadata?

tetsuo-cpp:
@artemdinaburg Are you familiar with this change, or should I leave it until @pgoodman returns.

libraries/anvill_passes/src/LowerRemillMemoryAccessIntrinsics.cpp
pgoodman: Yes, these new instructions should be tagged with the metadata of the `CallInst` which they replace.

libraries/anvill_passes/src/InstructionFolderPass.cpp
```diff
@@ -464,8 +465,8 @@ bool InstructionFolderPass::FoldSelectWithBinaryOp(
   auto new_true_value = builder.CreateBinOp(opcode, true_value, operand);
```
pgoodman, line 465: Need propagation here.

libraries/anvill_passes/src/InstructionFolderPass.cpp
```diff
@@ -464,8 +465,8 @@ bool InstructionFolderPass::FoldSelectWithBinaryOp(
...
   auto new_false_value = builder.CreateBinOp(opcode, false_value, operand);
```
pgoodman, line 466: Need propagation here.

libraries/anvill_passes/src/InstructionFolderPass.cpp
```diff
@@ -550,8 +552,8 @@ bool InstructionFolderPass::FoldSelectWithCastInst(
   auto new_false_value =
```
pgoodman, line 552: Need propagation here.

libraries/anvill_passes/src/InstructionFolderPass.cpp
```diff
@@ -550,8 +552,8 @@ bool InstructionFolderPass::FoldSelectWithCastInst(
...
       builder.CreateCast(cast_opcode, false_value, destination_type);
 
-  auto replacement =
-      builder.CreateSelect(condition, new_true_value, new_false_value);
+  auto replacement = builder.CreateSelect(condition, new_true_value,
```
pgoodman, line 555: Need propagation here.

libraries/anvill_passes/src/InstructionFolderPass.cpp
```diff
@@ -637,8 +640,8 @@ bool InstructionFolderPass::FoldSelectWithGEPInst(
     return false;
   }
 
-  auto replacement =
-      builder.CreateSelect(condition, new_true_value, new_false_value);
+  auto replacement = builder.CreateSelect(condition, new_true_value,
```
pgoodman, line 643: Need propagation here.

libraries/anvill_passes/src/LowerRemillMemoryAccessIntrinsics.cpp
```diff
@@ -66,7 +68,9 @@ static void ReplaceMemWriteOp(llvm::CallBase *call_inst, llvm::Type *val_type) {
     val = ir.CreateFPExt(val, val_type);
```
pgoodman, line 68: Need propagation here.

pgoodman: And above on the `CreateIntToPtr`

libraries/anvill_passes/src/LowerRemillMemoryAccessIntrinsics.cpp
```diff
@@ -50,6 +50,8 @@ static void ReplaceMemReadOp(llvm::CallBase *call_inst, llvm::Type *val_type) {
   if (val_type->isX86_FP80Ty() || val_type->isFP128Ty()) {
     val = ir.CreateFPTrunc(val, call_inst->getType());
```
pgoodman, line 51: Need propagation here.

libraries/anvill_passes/src/LowerRemillMemoryAccessIntrinsics.cpp
```diff
@@ -50,6 +50,8 @@ static void ReplaceMemReadOp(llvm::CallBase *call_inst, llvm::Type *val_type) {
...
   }
+  auto *val_inst = llvm::dyn_cast<llvm::Instruction>(val);
```
pgoodman, line 53: This propagation needs to be moved up to the `val = ir.CreateLoad`

pgoodman (changes requested):
I'd recommend making some kind of utility function that operates on two `llvm::Value *`s, then casts each to instructions, and if neither are `nullptr`, then does the metadata copying. Then you can call it like `CopyMetadataTo(src, dest)` or `CopyMetadataFrom(dest, src)`, and you don't need to worry about `dyn_cast` returning `nullptr` or `CHECK` failures. Basically, it can be opportunistic, and if we fail then that's not an error condition.

libraries/anvill_passes/src/InstructionFolderPass.cpp
```diff
@@ -263,7 +263,7 @@ bool InstructionFolderPass::FoldSelectInstruction(
 
     // Search for a function that knows how to handle this case
     auto instr_folder_it =
-        kSelectInstructionFolderMap.find(repl.original_instr->getOpcode());
+      kSelectInstructionFolderMap.find(repl.original_instr->getOpcode());
```
tetsuo-cpp, line 266: This was Clang Format.

tetsuo-cpp:
@pgoodman 
I realised I missed a lot of spots so I've fixed those too + addressed the lines you pointed out.

pgoodman (approved):

PR #252, opened 2021-08-19
Migrate to the new LLVM pass manager

### LINKED ISSUE #130: Migrating to the new LLVM pass manager
We're widely using `llvm::legacy` pass managers. Might be a good idea to migrate eventually.

[How to write a new LLVM pass](https://llvm.org/docs/WritingAnLLVMNewPMPass.html)
[Using the new pass manager](https://llvm.org/docs/NewPassManager.html)

tetsuo-cpp:
Closes #130

### DIFF
--- a/anvill/src/Optimize.cpp
+++ b/anvill/src/Optimize.cpp
@@ -31,11 +31,24 @@
 #include <llvm/IR/InlineAsm.h>
 #include <llvm/IR/Instruction.h>
 #include <llvm/IR/Instructions.h>
-#include <llvm/IR/LegacyPassManager.h>
 #include <llvm/IR/Module.h>
 #include <llvm/IR/Type.h>
+#include <llvm/Passes/PassBuilder.h>
 #include <llvm/Transforms/InstCombine/InstCombine.h>
+#include <llvm/Transforms/Scalar/DCE.h>
+#include <llvm/Transforms/Scalar/Sink.h>
+#include <llvm/Transforms/Scalar/NewGVN.h>
+#include <llvm/Transforms/Scalar/SCCP.h>
+#include <llvm/Transforms/Scalar/DeadStoreElimination.h>
+#include <llvm/Transforms/Scalar/SROA.h>
+#include <llvm/Transforms/Scalar/EarlyCSE.h>
+#include <llvm/Transforms/Scalar/BDCE.h>
+#include <llvm/Transforms/Scalar/SimplifyCFG.h>
 #include <llvm/Transforms/IPO.h>
+#include <llvm/Transforms/IPO/Inliner.h>
+#include <llvm/Transforms/IPO/GlobalOpt.h>
+#include <llvm/Transforms/IPO/GlobalDCE.h>
+// #include <llvm/Transforms/IPO/StripSymbols.h>
 #include <llvm/Transforms/Utils/Local.h>
 
 // clang-format on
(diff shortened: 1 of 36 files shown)


tetsuo-cpp:
The unit tests seem ok but if I run `anvill-decompile`, I'm seeing some ASan failures.

anvill/src/Optimize.cpp
```diff
@@ -84,63 +97,76 @@ void OptimizeModule(const EntityLifter &lifter_context,
...
+
+  llvm::ModulePassManager mpm;
+  llvm::InlineParams inline_params;
+  inline_params.DefaultThreshold = 250;
+  mpm.addPass(llvm::ModuleInlinerWrapperPass(inline_params));
+  mpm.addPass(llvm::GlobalOptPass());
+  mpm.addPass(llvm::GlobalDCEPass());
+  // TODO(alex): Figure out what to do here.
+  // The new version of this pass only seems to have been added in LLVM 12.
+  // mpm.addPass(llvm::StripDeadDebugInfoPass());
```
tetsuo-cpp, line 119: This pass was introduced in [this commit](https://github.com/llvm/llvm-project/commit/ba12e77ec16b38a4498610c6b8cdeb1a7e8a6aae). It'd be pretty easy to just write the wrapper ourselves if we wanted.

libraries/anvill_passes/src/TransformRemillJumpIntrinsics.cpp
```diff
@@ -202,17 +204,17 @@ bool TransformRemillJumpIntrinsics::runOnFunction(llvm::Function &func) {
...
-    fpm.doInitialization();
-    fpm.run(func);
-    fpm.doFinalization();
+    llvm::FunctionPassManager fpm;
+    fpm.addPass(llvm::DCEPass());
+    fpm.addPass(llvm::SROA());
+    fpm.addPass(llvm::SimplifyCFGPass());
+    // TODO(alex): It doesn't seem to like this and complains about the pass not being registered.
+    // Perhaps it's because we're nesting pass managers?
+    // fpm.addPass(llvm::InstCombinePass());
```
tetsuo-cpp, line 213: This thing where we do another series of passes in a nested way seems to be causing a bit of trouble.

anvill/src/Optimize.cpp
```diff
@@ -84,63 +97,76 @@ void OptimizeModule(const EntityLifter &lifter_context,
...
-  fpm.add(llvm::createDeadCodeEliminationPass());
-  fpm.add(CreateRecoverEntityUseInformation(err_man, lifter_context));
-  fpm.add(CreateSinkSelectionsIntoBranchTargets(err_man));
-  fpm.add(CreateRemoveTrivialPhisAndSelects());
-  fpm.add(llvm::createDeadCodeEliminationPass());
-  fpm.add(CreateRecoverStackFrameInformation(err_man, options));
-  fpm.add(llvm::createSROAPass());
-  fpm.add(CreateSplitStackFrameAtReturnAddress(err_man));
-  fpm.add(llvm::createSROAPass());
+  AddSinkSelectionsIntoBranchTargets(fpm, err_man);
```
tetsuo-cpp, line 139: The new pass manager doesn't take a pointer anymore. So if I want to do:
```
fpm.add(SinkSelectionsIntoBranchTargets(err_man)
```
Then I'd need to expose the headers for each of the passes and move them to `libraries/anvill_passes/include` since we need the actual definitions to construct it. To avoid reorganising everything, I've exposed these helpers in `Transform.h` to add the pass to the pass manager.

PR #258, opened 2021-09-14
Handle i128 arg types for arm64

kumarak:
Ignore `xzr` register for arm64 and unsupported registers in x86 binaries. Handle i128 parameter types in arm64.

### DIFF
--- a/anvill/python/anvill/binja/bnfunction.py
+++ b/anvill/python/anvill/binja/bnfunction.py
@@ -28,6 +28,14 @@ from .xreftype import *
 from anvill.function import *
 from anvill.type import *
 
+def should_ignore_register(bv, reg_name):
+    _IGNORE_REGS_LIST = {
+        "aarch64": ["XZR", "WZR"]
+    }
+    try:
+        return (reg_name.upper() in _IGNORE_REGS_LIST[bv.arch.name])
+    except KeyError:
+        return False
 
 class BNFunction(Function):
     def __init__(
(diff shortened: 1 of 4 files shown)


anvill/python/anvill/binja/bnfunction.py
```diff
@@ -152,6 +163,9 @@ def _extract_types(
         rdi, rsi as operands in the MLIL, we should check if they have pointer information)
         """
         results = []
+        if not (item_or_list and str(item_or_list)):
```
pgoodman, line 166: What is going on here with the `str`?

anvill/python/anvill/binja/bnfunction.py
```diff
@@ -198,7 +215,7 @@ def _fill_bytes(self, program, memory, start, end, ref_eas):
                 seg = program.bv.get_segment_at(ea)
                 br.seek(ea)
 
-                #NOTE(artem): This is a workaround for binary ninja's fake
+                # NOTE(artem): This is a workaround for binary ninja's fake
```
pgoodman, line 218: Thanks :-)

pgoodman (approved):

anvill/python/anvill/binja/bnfunction.py
kumarak: `item_or_list` may not be an empty object but it's string representation may be empty. In that case there is possibility that register name is empty for unimplemented operands.

pgoodman: Huh, can you add a comment for this. That is rather subtle.

PR #259, opened 2021-09-21
wip: enable caching instructions in xref resolver

kumarak:
Enabled caching of instruction in xref resolver to avoid exploding the states while doing the lookup and resolving cross references. The caching is also added while looking for the value derived from symbolic stack.

### DIFF
--- a/anvill/src/Analysis/CrossReferenceResolver.cpp
+++ b/anvill/src/Analysis/CrossReferenceResolver.cpp
@@ -15,14 +15,12 @@
  * along with this program.  If not, see <http://www.gnu.org/licenses/>.
  */
 
-#include <glog/logging.h>
-
 #include <anvill/ABI.h>
 #include <anvill/Analysis/CrossReferenceResolver.h>
 #include <anvill/Analysis/Utils.h>
 #include <anvill/Lifters/EntityLifter.h>
 #include <anvill/Lifters/Options.h>
-
+#include <glog/logging.h>
 #include <llvm/IR/Constant.h>
 #include <llvm/IR/Constants.h>
 #include <llvm/IR/DataLayout.h>
@@ -251,9 +249,11 @@ ResolvedCrossReference CrossReferenceResolverImpl::MergeLeft(
 
 ResolvedCrossReference
 CrossReferenceResolverImpl::ResolveInstruction(llvm::Instruction *inst_val) {
+  auto &xr = xref_cache[inst_val];
 
   auto opnd_type = inst_val->getOperand(0)->getType();
   uint64_t size = opnd_type->getPrimitiveSizeInBits();
+
   // update size if the operand is of pointer type
   if (opnd_type->isPointerTy()) {
     size = dl.getPointerSizeInBits(0);
@@ -269,8 +269,8 @@ CrossReferenceResolverImpl::ResolveInstruction(llvm::Instruction *inst_val) {
   switch (inst_val->getOpcode()) {
 #define FOLD_CASE(name) \
   case llvm::Instruction::name: { \
-    auto xr = Fold##name(ResolveValue(inst_val->getOperand(0)), \
-                         ResolveValue(inst_val->getOperand(1)), mask, size); \
+    xr = Fold##name(ResolveValue(inst_val->getOperand(0)), \
+                    ResolveValue(inst_val->getOperand(1)), mask, size); \
     xr.size = static_cast<unsigned>(out_size); \
     return xr; \
   }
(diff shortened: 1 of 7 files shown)


anvill/src/Analysis/CrossReferenceResolver.cpp
```diff
@@ -251,9 +249,11 @@ ResolvedCrossReference CrossReferenceResolverImpl::MergeLeft(
 
 ResolvedCrossReference
 CrossReferenceResolverImpl::ResolveInstruction(llvm::Instruction *inst_val) {
+  auto &xr = xref_cache[inst_val];
```
pgoodman, line 252: Check if `xr` is present, or that `xr.flat` is non-zero, and early return?

anvill/src/Analysis/CrossReferenceResolver.cpp
```diff
@@ -353,12 +353,6 @@ CrossReferenceResolverImpl::ResolveInstruction(llvm::Instruction *inst_val) {
 // Try to resolve a constant to a cross-reference.
 ResolvedCrossReference
 CrossReferenceResolverImpl::ResolveConstant(llvm::Constant *const_val) {
-
-  auto it = xref_cache.find(const_val);
```
pgoodman, line 357: Why disabling of the fast path of the cache here?

anvill/src/Analysis/CrossReferenceResolver.cpp
```diff
@@ -650,12 +646,16 @@ CrossReferenceResolverImpl::ResolveCall(llvm::CallInst *call) {
 // Try to resolve `val` as a cross-reference.
 ResolvedCrossReference
 CrossReferenceResolverImpl::ResolveValue(llvm::Value *val) {
+  auto it = xref_cache.find(val);
```
pgoodman, line 649: OK I think I understand; at a higher level you check the cache; though when we're resolving constants, we always recursively call `ResolveConstant`, so now we're losing out there. I think we want to move this back into `ResolveConstant` and `ResolveInstruction`.

anvill/src/Analysis/CrossReferenceResolver.cpp
```diff
@@ -698,9 +698,18 @@ void CrossReferenceResolver::ClearCache(void) const {
...
+// if the application is using the cache and does not want to invalidate it.
 ResolvedCrossReference
-CrossReferenceResolver::TryResolveReference(llvm::Value *val) const {
+CrossReferenceResolver::TryResolveReference(llvm::Value *val,
+                                            bool uses_cache) const {
+
+  // If the application is not using cache, invalidate it before resolving
+  // the cross references. It is done to avoid stale `val` sitting in the
+  // cache if it has been changed/deleted.
+  if (!uses_cache) {
```
pgoodman, line 710: I don't like this `bool` parameter; I think perhaps what you could do is instead rename the method, and add a method:

```
TryResolveReferenceWithCaching
TryResolveReferenceWithClearedCache
```

Or something like this. That, or pass in an explicit cache datastructure, and then maybe `xref_cache` a pointer to it, that you can set/reset. E.g.:

```
using CrossReferenceCache = std::unordered_map<llvm::Value *, ResolvedCrossReference>;
```

```
CrossReferenceCache *xref_cache{nullptr};

...

ResolvedCrossReference
CrossReferenceResolver::TryResolveReference(llvm::Value *val,
                                            CrossReferenceCache &xref_cache_) const {
  impl->xref_cache = &xref_cache_;
  auto xr = impl->ResolveValue(val);
  impl->xref_cache = nullpltr;
  return xr;
}
```

Then above you do things like:
```
ResolvedCrossReference dummy_xr;
ResolvedCrossReference *cached_xr = &dummy_xr;
if (xref_cache) {
  auto it = xref_cache->find(val);
  if (it != xref_cache->end()) {
    return it->second;
  }
  cached_xr = &(xref_cache->emplace(val, dummy_xr).first->second);
}
auto &xr = *cached_xr;
```

pgoodman (changes requested):

anvill/src/Analysis/CrossReferenceResolver.cpp
kumarak: The fast path is moved to `ResolveValue` and may not be needed here. But there is no harm having it here as well for cases where `ResolveConstant` is getting called directly.

pgoodman (approved):

anvill/src/Analysis/CrossReferenceResolver.cpp
kumarak: I updated the function name to:
```
      TryResolveReferenceWithCaching
       TryResolveReferenceWithClearedCache
```
It will be nice for `CrossReferenceResolver` to always use cache internally if it save analysis time. If the application wants to access the cache, we can add an interface function to expose it.

pgoodman (approved):

Issue #219, opened 2021-06-10
Detect correct subarchitecture of amd64 in Python


artemdinaburg:
Right now we always use the generic `amd64` architecture for all amd64 binaries. However, some binaries will use ZMM registers for AVX instructions.  Lifting these will then fail, since we need to specify the `amd64_avx`/`amd64_avx512` architectures if the target uses ymm/zmm registers, respectively.

The Python code should be able to identify if any avx/avx512 registers have been encountered, and, if so, specify the correct subarchitecture.

artemdinaburg:
Adding a note on this and a bump in priority:
We also need to handle it when x86 includes things like avx/avx2 instructions. These are technically valid encodings in x86, but will only run on CPUs that also support amd64. Currently our x86 implementation assumes older CPU that is 32-bit, not a newer CPU that can run in 32-bit mode, and hence there is no avx/avx2 support in the x86 model.

We can use similar infrastructure to detect functions that use avx/avx2 and mark these as "do not want" for lifting.

PR #260, opened 2021-10-01
Detect sub architecture of amd64 in python

kumarak:
The changes scan the registers and detect correct sub-architecture if xmm/ymm/zmm registers getting used. (#219)

### DIFF
--- a/anvill/python/anvill/arch.py
+++ b/anvill/python/anvill/arch.py
@@ -370,13 +370,13 @@ class AMD64Arch(Arch):
 
     def register_name(self, reg_name) -> Register:
         if reg_name.startswith("%"):
-            ret_name = reg_name[1:].upper()
+            reg_name = reg_name[1:].upper()
         else:
-            ret_name = reg_name.upper()
+            reg_name = reg_name.upper()
 
-        if ret_name.startswith("ZMM"):
+        if reg_name.startswith("ZMM"):
             self._has_avx512 = True
-        elif ret_name.startswith("YMM"):
+        elif reg_name.startswith("YMM"):
             self._has_avx = True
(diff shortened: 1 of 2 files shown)


anvill/python/anvill/arch.py
```diff
@@ -361,6 +369,8 @@ def register_family(self, reg_name) -> Tuple[RegisterBounds, ...]:
         return self._REG_FAMILY[self.register_name(reg_name)]
 
     def register_name(self, reg_name) -> Register:
+        self._has_avx |= any(reg in reg_name for reg in ["xmm", "ymm", "XMM", "YMM"])
```
pgoodman, line 372: Change to the following. I'm not a fan of `any` over the list allocation that always needs to happen.

```python
        if reg_name.startswith("%"):
            ret_name = reg_name[1:].upper()
        else:
            ret_name = reg_name.upper()

        if ret_name.startswith("ZMM"):
            self._has_avx512 = True
        elif ret_name.startswith("YMM"):
            self._has_avx = True

        return ret_name
```

anvill/python/anvill/arch.py
```diff
@@ -606,6 +624,8 @@ def register_family(self, reg_name) -> Tuple[RegisterBounds, ...]:
...
```
pgoodman, line 627: Copy from above.

pgoodman (changes requested):

anvill/python/anvill/arch.py
kumarak: I updated the checking for `YMM` and `XMM` registers.

pgoodman (approved):

PR #261, opened 2021-10-05
Ian/switch lowering

2over12:
This PR implements recovery of switch statements to remove uses of the anvill_complete_switch intrinsic. The goal is to recover switch implementations that obey the following constraints:
1. The program counter depends on only constants and a load from memory.
2. This load only depends on one non-constant value.
3. This non-constant value is bounded by a comparison that is the terminator of the immediate dominator of the indirect jump.

The algorithm recovers switches by slicing the program counter to the target load. The load's address is then sliced to a candidate index. The constraint extractor then extracts the constraints from the branch that restrict the index in the block where the indirect jump occurs. If the constraint extractor encounters a cast and the cast is on the definition chain for the load address then the index is updated to that cast because precise bounds on casts do not always exist. Once the constraints are collected, the jump table analysis passes off the constraints to z3, attempting to prove a conservative range on the index. The solver selects the narrowest range between a signed and unsigned range. The program counter slice, index slice, and bounds are stored as a jump table result.

Transforming anvill_complete_switch intrinsics is then a matter of reading off values from the jump table. The switch lowering pass builds a mapping from the original switch called pcbinding. pcbinding maps the program counter to a target label. The switch lowering pass then inserts cases into a new switch by iterating from the lower bound on the index to the upper bound of the index. Each possible index value is interpreted in the index slice to retrieve a jump table address. The offset is then loaded from the jump table and the program counter slice is interpreted with this loaded value. If the memory is unavailable the switch is not lowered. This final program counter value is used to lookup the corresponding label for the initial index.

The PR adds two passes: JumpTableAnalysis and SwitchLoweringPass. The jump table analysis recovers the target slices  and constraints. The switch lowering pass depends on the jump table analysis and transforms anvill_complete_switch intrinsics to llvm switch statements.

Note:
An analysis that is providing memory to the SwitchLoweringPass must be careful. The conservative bound is not guaranteed to be small and could indeed point to all of memory. A providing analysis could provide sampled random portions of the bound to see if the loaded value results in a valid program counter in pcbinding so that over-conservative bounds can be filtered.

Currently, the comparison and index computations for the jumptable are left dead, but in the future these should be removed because the behavior is subsumed by the recovered switch construct.


libraries/anvill_passes/include/anvill/IndirectJumpPass.h
```diff
@@ -0,0 +1,59 @@
...
+  if (const auto *callee = callinsn->getCalledFunction()) {
+    return callee->getName().equals(kAnvillSwitchCompleteFunc);
+  }
+
+  return false;
+}
+
+std::vector<llvm::CallInst *> getTargetCalls(llvm::Function &F) {
+  std::vector<llvm::CallInst *> calls;
+  for (auto &blk : F.getBasicBlockList()) {
```
pgoodman, line 19: Two alternate ways of writing this:

```c++
for (auto &blk : *F) {
  for (auto &insn : blk) {
    ...
  }
}
```

The other is:

```c++
for (auto &insn : llvm::instructions(*F)) {
  ...
}
```

While it is common in the LLVM codebase to use capitalized variables like `F`, in this codebase use all lower-case variables names.

libraries/anvill_passes/include/anvill/IndirectJumpPass.h
```diff
@@ -0,0 +1,59 @@
+#pragma once
+
+#include <anvill/ABI.h>
+#include <llvm/IR/Instructions.h>
+#include <llvm/Pass.h>
+
+namespace anvill {
+namespace {
+bool isTargetInstrinsic(const llvm::CallInst *callinsn) {
```
pgoodman, line 9: When putting things in anonymous namespaces, I like to also use `static` (even though redundant) to really clarify that this function / variable will end up local to the translation unit into which it is included.

libraries/anvill_passes/include/anvill/IndirectJumpPass.h
```diff
@@ -0,0 +1,59 @@
...
+}
+}  // namespace
+
+// (ian) Unfortunately pretty sure CRTP is the only way to do this without running into issues with pass IDs
+template <typename UserFunctionPass>
+class IndirectJumpPass : public llvm::FunctionPass {
+ public:
+  static char ID;
+
+  IndirectJumpPass() : llvm::FunctionPass(ID) {}
```
pgoodman, line 40: `IndirectJumpPass(void) : ...`

libraries/anvill_passes/include/anvill/SliceManager.h
```diff
@@ -0,0 +1,107 @@
...
+                             llvm::Value *returnVal);
+
+
+ public:
+  static std::string getFunctionName(SliceID id) {
+    return "sliceFunc." + std::to_string(id.ID);
+  }
+
+  /**
+             * @brief Adds a slice of instructions to the slice manager. Any values not defined in the slice are lifted to arguments.
```
pgoodman, line 84: Why is this so indented? Perhaps the result of clang-format, or a side-effect my random post-processing script?

pgoodman: We generally just end up using `//` comments everywhere.

libraries/anvill_passes/src/JumpTableAnalysis.cpp
```diff
@@ -0,0 +1,639 @@
+#include <anvill/ABI.h>
```
pgoodman, line 1: Can you add some more comments into this code that describes what / why it's doing things, and perhaps have a file-level comment that describes the approach at a high-level?

libraries/anvill_passes/src/JumpTableAnalysis.cpp
```diff
@@ -0,0 +1,639 @@
...
+        return std::nullopt;
+      }
+
+      const llvm::Instruction *firstCTIInsns = targetCTIBlock->getFirstNonPHI();
+      const llvm::SmallPtrSet<llvm::BasicBlock *, 1> checkSet{
+          branch->getParent()};
+      const llvm::SmallPtrSetImpl<llvm::BasicBlock *> *st = &checkSet;
+
+      // newer versions of llvm let you go from blocks...
+      // TODO should pass loop info
```
pgoodman, line 454: format for todos: `TODO(ian): ...`

pgoodman (changes requested):

pgoodman:
Can you copy some of the paragraphs of text that you wrote in this PR into the code somewhere, so that the code documents at a high level the algorithm / approach?

pgoodman (changes requested):

libraries/anvill_passes/include/anvill/IndirectJumpPass.h
```diff
@@ -0,0 +1,58 @@
...
+#include <anvill/ABI.h>
+#include <llvm/IR/InstIterator.h>
+#include <llvm/IR/Instructions.h>
+#include <llvm/Pass.h>
+
+namespace anvill {
+namespace {
+static bool isTargetInstrinsic(const llvm::CallInst *callinsn) {
+  if (const auto *callee = callinsn->getCalledFunction()) {
+    return callee->getName().equals(kAnvillSwitchCompleteFunc);
```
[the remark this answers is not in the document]
2over12: So this is an interesting question and I meant to bring it up in the PR. Yes because the range for the index is conservative, the analysis will always consider all possible loads from the offset table. Now if the switch is truly incomplete (ie. we were not able to recover all possible jump targets) then the pcbinding lookup in switch lowering should fail because we won't have a label for one of the outgoing program counters. So in that sense I think the switch recovery pass (the actual switch generation) will only succeed for mislabeled incomplete switch intrinsics which then begs the question, should we even attempt to run that pass on the incomplete intrinsic? That being said the JumpTableAnalysis results will still be valid, even if the intrinsic is incomplete. Perhaps the best solution is to apply the analysis to both types of intrinsics and only run the lowering pass for complete switches? The assumption being that for incomplete switches we need to pass back that jump table analysis to the lifter itself.

pgoodman:
@2over12 when you have time, can you resolve conflicts and re-submit?

PR #262, opened 2021-10-06
Test on LLVM 12

ekilmer:
Looks like there's a failing test on LLVM 12


ekilmer:
Looks like the issue is with the executable name:

`anvill-decompile-json-12` is installed here https://github.com/lifting-bits/anvill/pull/262/checks?check_run_id=3816450561#step:17:16

but the integration test is trying to execute `anvill-decompile-json-12.0` here https://github.com/lifting-bits/anvill/pull/262/checks?check_run_id=3816450561#step:24:15

PR #270, opened 2021-11-09
This PR simplifies some APIs, such as type specification stuff and co…

pgoodman:
…ntrol-flow provider to remove things like result/error types that aren't practically useful. This PR also makes the type encoder/decoder use a single instance that can be used for multiple things/times. There was also a bug fix in the type decoder related to alphanumeric types. Finally, the control-flow provider and type provider are extended to be able to make more contextually-sensitive requests, e.g. the type of a function, given its call site.


artemdinaburg:
Closing since this is subsumed by the API simplification & branch condition combined PR.

2over12:
Closing in favor of #274

PR #290, opened 2022-03-04
Fix anvill commiting to default function types for all pointers

2over12:
This PR allows for some fairly ugly cooperation between the type provider and value lifter to avoid committing to default types for pointers without entities. In many cases, we use anvill to lift portions of a binary without lifting all of the entities. In this case the type provider is used to hint about whether we have knowledge about what exists at a given variable. In the presence of type providers that just provide a default address for a type this doesnt work, so the type provider now hints that decl it's returning doesnt imply any knowledge about this specific EA

### DIFF
--- a/include/anvill/Declarations.h
+++ b/include/anvill/Declarations.h
@@ -108,8 +108,8 @@ struct VariableDecl {
   std::uint64_t address{0};
 
   // Declare this global variable in an LLVM module.
-  llvm::GlobalVariable *DeclareInModule(
-      const std::string &name, llvm::Module &) const;
+  llvm::GlobalVariable *DeclareInModule(const std::string &name,
+                                        llvm::Module &) const;
 };
 
 // A declaration for a callable entity.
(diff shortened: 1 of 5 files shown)


include/anvill/Declarations.h
```diff
@@ -160,6 +160,10 @@ struct CallableDecl {
   // is a variadic function.
   bool is_variadic{false};
 
+
+  // This value provides a way for type providers to hint to lifters wether this type can be used to infer an entity is going to exist at this location.
+  bool is_default_decl{false};
```
pgoodman, line 165: I'm not exactly a fan of this. My preference is on of the following:
 1) make this a descriptive enum, where you can enumerate more than two cases we already have, e.g. "known function, known type", "known function, unknown type", "unknown function, unknown type."
 2) Make the same enum, but pass it as a parameter down through the type provider calls.

My preference is the latter, so that you could have a wrapping type provider change the confidence in the decl for wrapped providers, without altering the decl itself.

2over12: I dont see how option 2 could possibly work. We would need to pass down to the type provider some hint about wether we care about existence of the entity for which the type is being provided, then each type provider needs to have an internal notion of what it means for an entity to exist in its view. 

As for option 1 what other cases exist. We already know wether the type is known or not based on if we returned one. The case we are trying to disambiguate here is wether this type implies that an entity will exist there in the future (or at least can exist there in the future). Realistically, this indicates to me that we are conflating tasks in the type provider somehow. ie. we shouldnt ever really provide a guessed spec for a pointer because the calling context is a much better guess (or it's not even a function)

2over12: Really I guess perhaps what we want is in the value lifter in get pointer we want to know if this pointer is a global or a function then we should only call GetFunction or GetVariable type, not both

pgoodman: I see your point on (2), can you encode more of your comments above (the contextual reasoning) into a code comment and use an enum? I think distinguishing the three cases is useful and helps to explain the context, even if you end up always checking for just one of the three cases.

pgoodman (approved):

PR #291, opened 2022-03-14
Uses Default Versions of Type Provider Functions when Needed

2over12:
Uses the explicit default version of the type provider when defaults are acceptable.

### DIFF
--- a/include/anvill/Providers.h
+++ b/include/anvill/Providers.h
@@ -41,13 +41,8 @@ class TypeProvider {
   virtual std::optional<VariableDecl>
   GetDefaultVariableDecl(uint64_t address) const;
 
-
   std::optional<FunctionDecl>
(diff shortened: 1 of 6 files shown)


2over12:
Alright so the trouble with state escape is we get a store of the called function into RIP. The hinted type for that pointer is going to be an i8* because of the types within the state. Currently, the value lifter takes the hinted type over a default type with the assumption being that a hinted function call might have better context, but I dont think that is ever the case. I am going to change the code so that we take a default instead of the hint if the hint is not a function type and we know there is a function there. Really, I think the type provider is overloaded with functionality at this point. It is awkward and error prone to rely on the type provider to hold knowledge of what entities exist

pgoodman (approved):

pgoodman (approved):

PR #296, opened 2022-05-18
Support opaque pointers

tetsuo-cpp:
- [x] Fix unit tests


lib/CrossReferenceFolder.cpp
```diff
@@ -335,10 +335,11 @@ CrossReferenceFolderImpl::ResolveInstruction(llvm::Instruction *inst_val) {
     case llvm::Instruction::IntToPtr: {
       xr = ResolveValue(inst_val->getOperand(0));
       xr.size = static_cast<unsigned>(out_size);
-      if (auto ptr_type = llvm::cast<llvm::PointerType>(inst_val->getType());
-          !xr.displacement_from_hinted_value_type) {
-        xr.hinted_value_type = ptr_type->getElementType();
-      }
+      // NOTE(alex): Looks like this just improves the fidelity of the lift but not correctness?
```
tetsuo-cpp, line 338: This is the case for a lot of blocks I've commented out. I'd be interested to know which ones are reasonable to remove and which ones need to remain (and have the underlying pointer type propagated somehow).

pgoodman: I'd liek to see what happens if you delete `hinted_value_type` altogether 😈

pgoodman: And as a consequence, anything reliant on it.

tetsuo-cpp: > I'd like to see what happens if you delete `hinted_value_type` altogether 😈

I'll give that a try.

tetsuo-cpp: [6e4821e](https://github.com/lifting-bits/anvill/pull/296/commits/6e4821e3587b60892bc5c3c0b999874d9a1c5151)

lib/Lifters/ValueLifter.cpp
```diff
@@ -258,15 +243,12 @@ llvm::Constant *ValueLifterImpl::GetPointer(uint64_t ea, llvm::Type *value_type,
...
-
   std::stringstream ss;
-  ss << kGlobalAliasNamePrefix << std::hex << ea << '_'
-     << type_specifier.EncodeToString(type,
-                                      EncodingFormat::kValidSymbolCharsOnly);
+  ss << kGlobalAliasNamePrefix << std::hex << ea << '_';
 
   const auto name = ss.str();
-  auto alias_ret = llvm::GlobalAlias::create(value_type, address_space,
+  // TODO(alex): Might need the pointer type here. Try to understand what this is doing later.
```
tetsuo-cpp, line 250: I'm unsure about this spot.

tetsuo-cpp:
@pgoodman By the way, do we want to retain backwards compatibility with older LLVM versions? If so, we need to wrap most of these calls in a function that has a bunch of `ifdef` based on the LLVM version.

pgoodman:
> @pgoodman By the way, do we want to retain backwards compatibility with older LLVM versions? If so, we need to wrap most of these calls in a function that has a bunch of `ifdef` based on the LLVM version.

Don't maintain backwards compatibility. Move to LLVM 14 and never look back!

tetsuo-cpp:
Ok, I think this is ready to look at more seriously. It's still hanging with LLVM 14 but it's not hitting asserts anymore so I believe once we figure out the hanging in DSE mentioned in https://github.com/lifting-bits/anvill/pull/297#issuecomment-1148868821, this will be able to complete a lift of `cat`.

lib/Arch/AArch32_C.cpp
```diff
@@ -124,18 +124,7 @@ AArch32_C::BindReturnValues(llvm::Function &function, bool &injected_sret,
...
-      value_declaration.type =
-          remill::NthArgument(&function, 0)->getType()->getPointerElementType();
-
-    } else if (function.hasParamAttribute(1, llvm::Attribute::StructRet)) {
-      value_declaration.type =
-          remill::NthArgument(&function, 1)->getType()->getPointerElementType();
-    }
-
-    value_declaration.type = llvm::PointerType::get(value_declaration.type, 0);
+    value_declaration.type = llvm::PointerType::get(function.getContext(), 0);
```
pgoodman, line 127: I feel like I am a bit iffy on this change, but I can't quite say why. Can you try to formulate an example on compiler explorer that uses `structret` (e.g. have a function return a big `struct` by value), and then show the llvm 12/13 and trunk llvm bitcode side-by-side, and then also make two json specs, one using the old build, one with the new build, and then see what bitcode we produce and how it relates to the original?

pgoodman: Can be for x86-64, doesn't need to be aarch32.

lib/Lifters/FunctionLifter.cpp
```diff
@@ -1322,7 +1321,7 @@ llvm::Function *FunctionLifter::GetOrDeclareFunction(const FunctionDecl &decl) {
 void FunctionLifter::AllocateAndInitializeStateStructure(
     llvm::BasicBlock *block) {
   llvm::IRBuilder<> ir(block);
-  const auto state_type = state_ptr_type->getElementType();
+  const auto state_type = llvm::StructType::getTypeByName(ir.getContext(), "struct.State");
```
pgoodman, line 1324: Use the architecture here.

lib/Lifters/ValueLifter.cpp
```diff
@@ -142,22 +131,14 @@ static llvm::Constant *UnwrapZeroIndices(llvm::Constant *ret,
 llvm::Constant *
 ValueLifterImpl::TryGetPointerForAddress(uint64_t ea,
                                          EntityLifterImpl &ent_lifter,
-                                         llvm::PointerType *hinted_type) const {
+                                         llvm::Type *hinted_type) const {
```
pgoodman, line 134: Gawd I wish that instead we refactored the value lifted into non-existence, and got it to use the `CrossReferenceResolver` somehow.

pgoodman (changes requested):

lib/Passes/Utils.cpp
```diff
@@ -194,7 +194,7 @@ llvm::Function *AddressOfReturnAddressFunction(llvm::Module *module) {
     case llvm::Triple::ArchType::x86_64:
     case llvm::Triple::ArchType::aarch64:
     case llvm::Triple::ArchType::aarch64_be:
-      func_name = "llvm.addressofreturnaddress.p0i8";
+      func_name = "llvm.addressofreturnaddress.p0";
```
pgoodman, line 197: Good find!

lib/Lifters/FunctionLifter.cpp
pgoodman: Hrmm, though I've asked you to mark the state structure type method as deprecated in remill. I guess undo that.. !

pgoodman (changes requested):

lib/Declarations.cpp
```diff
@@ -173,7 +173,7 @@ llvm::Value *CallableDecl::CallFromLiftedBlock(
   // Store the return address, and computed return stack pointer.
   ir.SetInsertPoint(block);
 
-  ir.CreateStore(ret_addr, remill::FindVarInFunction(block, "NEXT_PC"));
+  ir.CreateStore(ret_addr, remill::FindVarInFunction(block, "NEXT_PC").first);
```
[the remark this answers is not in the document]
tetsuo-cpp: [625b28a](https://github.com/lifting-bits/anvill/pull/296/commits/625b28ae4518fdb0ffd195f6179bea7e0ad429fb)

lib/Lifters/ValueLifter.cpp
tetsuo-cpp: Do want to do this in this PR? Or this more of a future "nice to have"?

lib/Arch/AArch32_C.cpp
tetsuo-cpp: I see you marked this as resolved. Do you still want me to do this?

lib/Lifters/ValueLifter.cpp
tetsuo-cpp: I'm going to explore removing this part like we discussed.

lib/Optimize.cpp
```diff
@@ -235,7 +235,8 @@ void OptimizeModule(const EntityLifter &lifter, llvm::Module &module) {
   // makes code easier to read and analyze. This is a fairly narrow optimization
   // but it comes up often enough for lifted code.
 
-  AddConvertIntegerToPointerOperations(fpm);
+  // NOTE(alex): Need to rewrite this pass to somehow not rely on typed pointers.
+  // AddConvertIntegerToPointerOperations(fpm);
```
tetsuo-cpp, line 239: This looks like a decent chunk of work. I'm thinking of trying to get this change in first and following up with a change to rewrite this pass.

Does this work? Or is this pass really integral to the working of Anvill.

lib/Lifters/ValueLifter.cpp
pgoodman: Nice to have, not this PR.

pgoodman (approved):
This looks good to merge to me, though it sounds like you're still exploring something.

lib/Optimize.cpp
tetsuo-cpp: 635b8200d902d168325ee81e298ac2c200fef0f7

lib/Lifters/ValueLifter.cpp
tetsuo-cpp: [fa59b3f](https://github.com/lifting-bits/anvill/pull/296/commits/fa59b3f7fac4f841ea571c3183c0cf2e147a759b)

Not sure if this is really fine. It doesn't cause problems with my `cat` program that I'm testing but I'm more just testing that the lift completes successfully.

pgoodman (approved):

PR #297, opened 2022-06-07
Support LLVM 14

tetsuo-cpp:
I need this in order to continue with opaque pointers. Basically, in LLVM 13 I can create opaque pointers but the API still returns typed pointer variables in some situations. Also some of the LLVM passes still haven't switched over to opaque pointers in 13 so they call `PointerType::getElementType` which asserts on IR that contains opaque pointers.


lib/Optimize.cpp
```diff
@@ -170,7 +171,7 @@ void OptimizeModule(const EntityLifter &lifter, llvm::Module &module) {
 
   fpm.addPass(llvm::SCCPPass());
   fpm.addPass(llvm::DSEPass());
-  fpm.addPass(llvm::SROA());
+  fpm.addPass(llvm::SROAPass());
```
tetsuo-cpp, line 174: Looks like this is an incompatible change. Are we happy for me to `ifdef` around this to support both 13 and 14?

tetsuo-cpp: @pgoodman mentioned that we'd be dropping support for older versions at some point. Do we want to do this now and only support 14+ going forwards?

tetsuo-cpp:
There seems to be some kind of regression when I build with LLVM 14. That example `cat` program that I'm lifting ends up hanging in `anvill-decompile-json`.

lib/Optimize.cpp
tetsuo-cpp: [ec5ab11](https://github.com/lifting-bits/anvill/pull/297/commits/ec5ab11ab20bf33b0c20cc9afabe89ac0a36f174)

pgoodman:
Is that our DSE pass (e.g. remill used to have one, though I thought we eliminated it), or is it an LLVM DSE pass?

tetsuo-cpp:
> Is that our DSE pass (e.g. remill used to have one, though I thought we eliminated it), or is it an LLVM DSE pass?

It's the LLVM DSE pass.

tetsuo-cpp:
As we discussed, here's the JSON that triggers the hang.
[cat.json.txt](https://github.com/lifting-bits/anvill/files/8855262/cat.json.txt)

artemdinaburg:
I just pushed the release button for remill since it now has LLVM14 support

tetsuo-cpp:
The runtime is dominated by `DSEPass` and `SinkingPass`. Within the actual DSE pass logic, the runtime seems to be split between a few chunks so it's not really clear what the issue is. Will continue drilling down.

Screen Shot 2022-06-08 at 9 45 16 pm

pgoodman:
I was not able to replicate the DSE hang, though I can see that it does spend time there. I think we could always just disable the pass. Besides that, going to merge.

pgoodman (approved):

tetsuo-cpp:
As we discussed, it's not really hanging it's just spending a LOT of time in DSE. I'm going to follow up with a PR to comment out the use of DSE since we're not 100% sure it's that useful and document this in an issue in case we ever need to reenable it.

PR #306, opened 2022-07-20
Fix backwards stack

frabert:
Should fix trailofbits/IRENE#99 but I haven't had the chance to test it yet, so it's a draft for now


frabert:
Haven't tested this in IRENE proper, but using Anvill this is working fine -- I think it's ready for merge

pgoodman (approved):

PR #311, opened 2022-08-15
Tracks Decoding Context for Instructions in Worklist

2over12:
This PR gets anvill building against remill https://github.com/lifting-bits/remill/pull/617 and should encompass most of the anvill side changes to support mixed mode. The last thing to work out here is how we want to specify building the initial context per function ie. for a thumb labeled function we need an initial context with TM set to 1. One possibility is to have a context assignment list in the spec for functions

### DIFF
--- a/lib/Lifters/FunctionLifter.cpp
+++ b/lib/Lifters/FunctionLifter.cpp
@@ -1055,8 +1055,8 @@ void FunctionLifter::VisitInstruction(
   // Even when something isn't supported or is invalid, we still lift
   // a call to a semantic, e.g.`INVALID_INSTRUCTION`, so we really want
   // to treat instruction lifting as an operation that can't fail.
-  (void) inst.GetLifter()->LiftIntoBlock(inst, block, state_ptr,
-                                         false /* is_delayed */);
+  std::ignore = inst.GetLifter()->LiftIntoBlock(inst, block, state_ptr,
+                                                false /* is_delayed */);
 
   // Figure out if we have to decode the subsequent instruction as a delayed
   // instruction.
--- a/lib/Lifters/FunctionLifter.cpp
+++ b/lib/Lifters/FunctionLifter.cpp
@@ -291,9 +291,9 @@ void FunctionLifter::VisitInvalid(const remill::Instruction &inst,
 // similarly to invalid instructions, with the exception that they can have
 // delay slots, and therefore the subsequent instruction may actually execute
 // prior to the error.
-void FunctionLifter::VisitError(const remill::Instruction &inst,
-                                remill::Instruction *delayed_inst,
-                                llvm::BasicBlock *block) {
+void FunctionLifter::VisitError(
+    const remill::Instruction &inst,
+    std::optional<remill::Instruction> &delayed_inst, llvm::BasicBlock *block) {
   VisitDelayedInstruction(inst, delayed_inst, block, true);
   MuteStateEscape(
       remill::AddTerminatingTailCall(block, intrinsics.error, intrinsics));
@@ -322,8 +322,8 @@ void FunctionLifter::VisitNoOp(
 // `inst.branch_taken_pc`. Execution thus needs to transfer to the instruction
 // (and thus `llvm::BasicBlock`) associated with `inst.branch_taken_pc`.
 void FunctionLifter::VisitDirectJump(
-    const remill::Instruction &inst, remill::Instruction *delayed_inst,
-    llvm::BasicBlock *block,
+    const remill::Instruction &inst,
+    std::optional<remill::Instruction> &delayed_inst, llvm::BasicBlock *block,
     const remill::DecodingContext::ContextMap &mapper) {
   VisitDelayedInstruction(inst, delayed_inst, block, true);
   llvm::BranchInst::Create(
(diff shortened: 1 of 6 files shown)


lib/Lifters/FunctionLifter.cpp
```diff
@@ -1023,15 +1055,15 @@ void FunctionLifter::VisitInstruction(remill::Instruction &inst,
   // Even when something isn't supported or is invalid, we still lift
   // a call to a semantic, e.g.`INVALID_INSTRUCTION`, so we really want
   // to treat instruction lifting as an operation that can't fail.
-  (void) inst_lifter->LiftIntoBlock(inst, block, state_ptr,
-                                    false /* is_delayed */);
+  (void) inst.GetLifter()->LiftIntoBlock(inst, block, state_ptr,
```
lkorenc, line 1058: `std::ignore`

lib/Lifters/FunctionLifter.cpp
```diff
@@ -1056,43 +1087,47 @@ void FunctionLifter::VisitInstruction(remill::Instruction &inst,
...
-      VisitConditionalFunctionReturn(inst, delayed_inst, block);
+      VisitConditionalFunctionReturn(inst, delayed_inst, block, mapper);
       break;
     case remill::Instruction::kCategoryDirectFunctionCall:
-      VisitDirectFunctionCall(inst, delayed_inst, block);
+      VisitDirectFunctionCall(inst, delayed_inst, block, mapper);
       break;
     case remill::Instruction::kCategoryConditionalDirectFunctionCall:
-      VisitDirectFunctionCall(inst, delayed_inst, block);
+      VisitDirectFunctionCall(inst, delayed_inst, block, mapper);
```
sschriner, line 1115: `VisitConditionalDirectFunctionCall`

lib/Lifters/FunctionLifter.cpp
```diff
@@ -1056,43 +1087,47 @@ void FunctionLifter::VisitInstruction(remill::Instruction &inst,
...
     case remill::Instruction::kCategoryIndirectJump:
-      VisitIndirectJump(inst, delayed_inst, block);
+      VisitIndirectJump(inst, delayed_inst, block, mapper);
       break;
     case remill::Instruction::kCategoryConditionalIndirectJump:
-      VisitConditionalIndirectJump(inst, delayed_inst, block);
+      VisitConditionalIndirectJump(inst, delayed_inst, block, mapper);
       break;
     case remill::Instruction::kCategoryFunctionReturn:
       VisitFunctionReturn(inst, delayed_inst, block);
```
sschriner, line 1106: Why does the function return and async hyper call not require the context?

2over12: Any purely interprocedural control flow does not need the context since no direct instruction lifting happens inside them. ie. for an asynchypercall the function terminates so there is nothing more to decode in this function. The reason we need it for say a condtionalasynchypercall is to handle decoding the fall-through

lib/Lifters/FunctionLifter.cpp
```diff
@@ -1056,43 +1087,47 @@ void FunctionLifter::VisitInstruction(remill::Instruction &inst,
     case remill::Instruction::kCategoryError:
       VisitError(inst, delayed_inst, block);
       break;
-    case remill::Instruction::kCategoryNormal: VisitNormal(inst, block); break;
-    case remill::Instruction::kCategoryNoOp: VisitNoOp(inst, block); break;
+    case remill::Instruction::kCategoryNormal:
```
lkorenc, line 1090: Extract as a standalone function/local lambda and replace `break` with `return`? It won't give that much gains though, but it should save a few line + make this function a bit more readable, which maybe is worth it?

lib/Lifters/FunctionLifter.cpp
lkorenc: Is it not worth to give all `VisitX` the same prototype?

lib/Lifters/FunctionLifter.cpp
```diff
@@ -843,9 +868,10 @@ void FunctionLifter::VisitConditionalBranch(const remill::Instruction &inst,
   VisitDelayedInstruction(inst, delayed_inst, taken_block, true);
   VisitDelayedInstruction(inst, delayed_inst, not_taken_block, false);
   auto br2 = llvm::BranchInst::Create(
```
lkorenc, line 870: Rename, what is `br2`? Where am I branching to? Same for others.

lkorenc (changes requested):

lib/Lifters/FunctionLifter.cpp
2over12: I'm not sure, we'd just be passing an unused parameter then and visits are not used uniformly anyways

lib/Lifters/FunctionLifter.cpp
```diff
@@ -1047,18 +1047,7 @@ void FunctionLifter::VisitInstruction(
...
-
-  // Reserve space for an instrucion that will go into a delay slot, in case it
-  // is needed. This is an uncommon case, so avoid instantiating a new
-  // Instruction unless it is actually needed. The instruction instantition into
-  // this buffer happens via a placement new call later on.
-  std::aligned_storage<sizeof(remill::Instruction),
-                       alignof(remill::Instruction)>::type delayed_inst_storage;
-
-  remill::Instruction *delayed_inst = nullptr;
+  std::unique_ptr<remill::Instruction> delayed_inst = nullptr;
```
lkorenc, line 1050: Why is this an `unique_ptr`? Before it was local memory - `std::optional` therefore seems on the first glance as more natural replacement? `std::unique_ptr` is heap memory, so of course if we want the data to survive for longer this is the way (but it was not before?)

lkorenc (changes requested):

lib/Lifters/FunctionLifter.cpp
2over12: yeah this probably isnt ideal... but the problem is all of the downstream methods take a pointer to an instruction then check if it's null, i suppose the correct thing here is to update every reference to the delayed instruction to be an optional

lkorenc (approved):

2over12:
Integration test failures are blocked on https://github.com/lifting-bits/remill/pull/620

Issue #310, opened 2022-08-15
Build fails with latest Remill


tetsuo-cpp:
With the latest Remill, I'm seeing compilation failures:
```
/Users/tetsuo/Code/anvill/lib/Lifters/FunctionLifter.cpp:161:7: error: no matching constructor for initialization of 'remill::InstructionLifter::LifterPtr' (aka 'shared_ptr<remill::InstructionLifter>')
      inst_lifter(options.arch->DefaultLifter(intrinsics)),
      ^           ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
/Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX12.3.sdk/usr/include/c++/v1/__memory/shared_ptr.h:441:23: note: candidate constructor not viable: no known conversion from 'OperandLifter::OpLifterPtr' (aka 'shared_ptr<remill::OperandLifter>') to 'std::nullptr_t' (aka 'nullptr_t') for 1st argument
    _LIBCPP_CONSTEXPR shared_ptr(nullptr_t) _NOEXCEPT;
```
I think Anvill requires some changes to work with https://github.com/lifting-bits/remill/pull/615.

2over12:
Fixed in #311

PR #312, opened 2022-08-21
Fix 310

Midi12:
Hello,

I also came across the compilation bug referenced by this issue [Build fails with latest Remill #310](https://github.com/lifting-bits/anvill/issues/310).

The pull request implements the new `remill::InstructionLifter` constructor prototype as well as handling of the new return type`std::optional< DecodingResult >` in `DecodeInstruction`and `DecodeDelayedInstruction`. I don't know well remill and anvil yet so I went the dumb way to handling the decoding result. If `DecodeInstructionInto` needs to return the new type as well I can update the PR.

Kind regards,
Midi


Midi12:
Just noticed the actual implementation came with #311. This can be closed.

PR #321, opened 2022-09-08
Ian/simplify redirections

2over12:
This PR simplifies redirections to only redirect targets, rather than instruction/call/jump sites. This narrows redirections role to only handle thunks. calls inside of a thunk when lifting can just behandled by treating the jump as a tail call naturally. This PR also cherry-picks a fix to avoid adding returns after certain errors


Ninja3047 (approved):
lgtm, lifting seems to work better for plt entries

PR #325, opened 2022-09-19
Make global constant if only readable bytes in gv

2over12:


Ninja3047 (approved):
LGTM, output seems to be improved

PR #326, opened 2022-09-19
Remove cruft

Ninja3047:
* Ignore variables with no references
   * no real point in lifting those
 * Append address to name (with the exception of externs/main) to avoid naming conflicts


2over12:
If binja didn't find a reference to a variable we may be able to fold it later so maybe we want to retain the ability behind a flag. Re: the rest looks fine, de-conflicting by address seems ok I guess, if we lift a library we are breaking some stuff tho. I'd like to not append addresses to exported symbols if possible

2over12 (approved):
LGTM

PR #327, opened 2022-09-24
Store type string alongside LLVM types in spec

frabert:


2over12:
Sorry can you re-cherry-pick your commit over latest

lib/JSON.cpp
```diff
@@ -361,6 +361,7 @@ JSONTranslator::DecodeValue(const llvm::json::Object *obj, const char *desc,
 
   if (auto maybe_type_str = obj->getString("type")) {
     std::string spec = maybe_type_str->str();
+    decl.type_string = spec;
```
2over12, line 364: I think you missed doing this assignment in ParseJsonIntoCallableDecl

2over12 (changes requested):

lib/JSON.cpp
frabert: Doesn't it use `DecodeValue` as well?

PR #328, opened 2022-09-28
Migrate to new spec format

frabert:


data_specifications/specification.proto
```diff
@@ -0,0 +1,195 @@
...
+  BT_FL96 = 20;
+  BT_FL128 = 21;
+  BT_M64 = 22;
+  BT_VOID = 23;
+  BT_PADDING = 24;
+}
+
+message TypeSpec {
+  message PointerType {
+    TypeSpec pointee = 1;
```
tetsuo-cpp, line 57: When we translate these to LLVM types, we're going to lose the pointee type. Do we need this?

include/anvill/Type.h
```diff
@@ -172,12 +233,8 @@ class TypeTranslator {
       llvm::Type *type,
       EncodingFormat alphanum = EncodingFormat::kDefault) const;
 
-  // Parse an encoded type string into its represented type.
-  //
-  // See `docs/TypeEncoding.md` for information on how different types are
-  // represented.
   Result<llvm::Type *, TypeSpecificationError>
-  DecodeFromString(const std::string_view str) const;
+  DecodeFromSpec(TypeSpec spec) const;
```
tetsuo-cpp, line 237: I think we also need a `TypeSpec EncodeToSpec(llvm::Type *type);`. I'll work on that now.

tests/tools/src/TypeSpecification.cpp
```diff
@@ -62,6 +63,7 @@ TEST_SUITE("TypeSpecifier") {
     }
   }
 
+  /*
```
tetsuo-cpp, line 66: This test is cumbersome to get compiling, let's save it for later.

data_specifications/specification.proto
frabert: It's useful for when we use the spec to guide Rellic

include/anvill/Type.h
frabert: Are we sure we actually need it?

tetsuo-cpp: Maybe not, looks like most of the calls are in the JSON generation part + unit tests. I'll try removing the JSON bits first.

tetsuo-cpp: Oh probably not. Anyway, I hadn't written that yet so all good.

tetsuo-cpp:
Currently figuring out what to do about the default callable spec.

lib/Protobuf.cpp
```diff
@@ -0,0 +1,422 @@
...
+      // appropriate place in a register / on the stack by the kernel.
+    } else {
+      decl.return_address.type = void_type;
+    }
+
+    /*
+    // Decode the value of the stack pointer on exit from the function, which is
+    // defined in terms of `reg + offset` for a value of a register `reg`
+    // on entry to the function.
+    if (auto ret_sp = obj->getObject("return_stack_pointer")) {
```
tetsuo-cpp, line 199: Does this need to be part of the spec format?

tetsuo-cpp:
@frabert This branch is building and the JSON translation unit and header have been removed. There are a number of methods which I've just stubbed out (mostly in the Protobuf translator), that I've stubbed out to get things building so our next goal should probably be to fill those in.

lib/Protobuf.cpp
frabert: @2over12

data_specifications/specification.proto
```diff
@@ -0,0 +1,195 @@
...
+
+message Function {
+  uint64 entry_address = 1;
+  Value return_address = 2;
+  repeated Parameter parameters = 3;
+  repeated Value returns = 4;
+  bool is_variadic = 5;
+  bool is_noreturn = 6;
+  CallingConvention calling_convention = 7;
+  optional TypeSpec type = 8;
```
tetsuo-cpp, line 135: I also added this back in. This was part of the old spec file format and is needed for some of the existing logic. I don't think it's documented at https://github.com/lifting-bits/anvill/blob/master/docs/SpecificationFormat.md but it is apparently used for external functions.

artemdinaburg:
What is the final binary that we are looking to get now? Is there one? It looks to me like this will deprecate `anvill-decompile-json`

artemdinaburg:
It looks like this actually builds and passes CI now! Would love a sanity check prior to merge (@frabert / @tetsuo-cpp ) :)

frabert:
LGTM!

.github/workflows/build.yml
```diff
@@ -376,19 +408,19 @@ jobs:
         run: |
           python3 -m pip install -r libraries/lifting-tools-ci/requirements.txt
           scripts/test-amp-challenge-bins.sh \
-            --python-cmd "python3 -m anvill" \
-            --decompile-cmd "anvill-decompile-json"
+          --ghidra-install-dir $GHIDRA_INSTALL_DIR \
+          --decompile-cmd "anvill-decompile-json"
```
Ninja3047, line 398: probably should change this to protobuf or something to avoid future confusion

.github/workflows/build.yml
```diff
@@ -311,12 +346,9 @@ jobs:
 
         run: |
           scripts/build-preset.sh debug
-
       - name: Install Binary Ninja
```
Ninja3047: no longer needed

include/anvill/Declarations.h
```diff
@@ -173,6 +157,15 @@ struct CallableDecl {
                       const remill::IntrinsicTable &intrinsics,
                       llvm::BasicBlock *block, llvm::Value *state_ptr,
                       llvm::Value *mem_ptr) const;
+
+  // Try to create a callable decl from a protobuf default callable decl
+  // specification. Returns a string error if something went wrong.
+  //
+  // NOTE(alex): This is following the same pattern as where we decode an entire
```
tetsuo-cpp, line 164: If you're ok with this static member function, we should get rid of this comment.

tests/tools/src/TypeSpecification.cpp
tetsuo-cpp: Still necessary.

include/anvill/Declarations.h
tetsuo-cpp: Whoops, I reviewed just as this got merged. I'll follow up with a PR to address these later.

Issue #303, opened 2022-07-06
Rewrite ConvertIntegerToPointerOperations pass to not rely on pointer types


tetsuo-cpp:
As part of the opaque pointers implementation, I commented out the `ConvertIntegerToPointerOperations` pass since it makes heavy use of pointer types. We should rewrite this pass to not rely on pointer types when we get the chance.

PR #330, opened 2022-10-04
Get Anvill building with LLVM 15

tetsuo-cpp:
The main change that might be unexpected here is the removal of the ConvertIntegerToPointerOperations pass.

We have an open issue (#303) to convert that pass to use opaque pointers and remove calls to `PointerType::getElementType`. With the introduction of LLVM 15, these calls have been removed entirely (they were deprecated in LLVM 14). So this pass no longer compiles.

Instead of fiddling around with the preprocessor to only compile this pass with LLVM 14, I figured it'd be best to just remove it since it's not even being added to the pass manager right now.


tetsuo-cpp:
Ok, the only remaining problem here is that the IR under `tests/anvill_passes/data` contain typed pointers and are incompatible with LLVM 15. So I'll need to find some way to regenerate these.

@pgoodman Do you know where these IR files came from?

pgoodman:
@tetsuo-cpp I think someone uses anvill to make them in the past, then just uploaded those files raw :-/

tetsuo-cpp:
> @tetsuo-cpp I think someone uses anvill to make them in the past, then just uploaded those files raw :-/

Yeah they're definitely generated by Anvill. But I was hoping that the executable sources (that got built with Clang and then lifted back to IR with Anvill) would exist somewhere.

Looks like maybe I'll just have to eyeball the IR to figure out what it's testing and see if I can replace them. I'll probably check in the new sources this time so we don't have the same problem again.

pgoodman (approved):

tetsuo-cpp:
@2over12 pointed out that I can regenerate the `.ll` files with LLVM's `opt` like so:
```console
opt --opaque-pointers ./anvill_passes/data/BaseFunctionPass.ll -S -o Test.ll
```
So I've gone ahead and done that. Unit tests seem to be passing so I think this is good to go.

pgoodman (approved):

PR #335, opened 2022-11-16
Ian/llvm 15 test fixes

2over12:
Rebased #330 for latest and fixes tests

### DIFF
--- a/tests/anvill_passes/src/Utils.cpp
+++ b/tests/anvill_passes/src/Utils.cpp
@@ -84,7 +84,10 @@ std::unique_ptr<llvm::LLVMContext> CreateContextWithOpaquePointers(void) {
   auto context = std::make_unique<llvm::LLVMContext>();
 #if LLVM_VERSION_NUMBER < LLVM_VERSION(15, 0)
   context->enableOpaquePointers();
+#else
+  context->setOpaquePointers(true);
 #endif
+
   return context;
 }
 


tests/anvill_passes/src/Utils.cpp
```diff
@@ -84,7 +84,10 @@ std::unique_ptr<llvm::LLVMContext> CreateContextWithOpaquePointers(void) {
   auto context = std::make_unique<llvm::LLVMContext>();
 #if LLVM_VERSION_NUMBER < LLVM_VERSION(15, 0)
   context->enableOpaquePointers();
+#else
+  context->setOpaquePointers(true);
```
tetsuo-cpp, line 88: Isn't this default behaviour? If not, this needs to be happen in a bunch more places in Anvill + my Remill change needs fixes too.

.github/workflows/build.yml
```diff
@@ -805,20 +804,3 @@ jobs:
       - name: Build LLVM ${{ matrix.llvm }} on ${{ matrix.ubuntu }}
         run: |
           docker build . -t docker.pkg.github.com/lifting-bits/anvill/anvill-llvm${{ matrix.llvm }}-ubuntu${{ matrix.ubuntu }}-amd64:latest -f Dockerfile --target=dist --build-arg REMILL_COMMIT_ID=${REMILL_COMMIT_ID} --build-arg UBUNTU_VERSION=${{ matrix.ubuntu }} --build-arg ARCH=amd64 --build-arg LLVM_VERSION=${{ matrix.llvm }}
-      - name: Test Docker image
```
tetsuo-cpp: This seems to be more a consequence of migrating to the Protobuf format right? Let's have that as a separate change.

tetsuo-cpp:
As per https://github.com/lifting-bits/anvill/pull/330#issuecomment-1297384225, my branch still needs more work to fix the testing. I'm going to change the merge target to my branch. By the way, feel free to just push to my branches in future.

tetsuo-cpp:
Sorry for the force pushing. I had to clean up my branch and aftewards, this patch no longer applied cleanly to it. This should be the correct diff though.

PR #339, opened 2022-11-28
Move to explicitly setting CMake toolchain file

tetsuo-cpp:


.remill_commit_id
```diff
@@ -1 +1 @@
-d5c50357287cae0072938477b3b525bafcce8069
\ No newline at end of file
+7f7d0255f296995044562700314832e2020778ef
```
tetsuo-cpp, line 1: Update this once https://github.com/lifting-bits/remill/pull/639 gets merged.

.remill_commit_id
```diff
@@ -1 +1 @@
-d5c50357287cae0072938477b3b525bafcce8069
\ No newline at end of file
+a8ead7b58496f0fc90100eee67de1eee74cdc8c3
```
ekilmer, line 1: Side-note, since we're here... Why don't we use a submodule?

CMakeLists.txt
```diff
@@ -23,7 +22,7 @@ endif(ANVILL_ENABLE_INSTALL)
 
 add_subdirectory("libraries")
 
-set(_VCPKG_SHARE_DIR "${VCPKG_ROOT}/installed/${VCPKG_TARGET_TRIPLET}/${CMAKE_INSTALL_DATADIR}")
+set(_VCPKG_SHARE_DIR "${Z_VCPKG_ROOT_DIR}/installed/${VCPKG_TARGET_TRIPLET}/${CMAKE_INSTALL_DATADIR}")
```
ekilmer, line 25: Hmmmm. This shouldn't be needed. I don't know the history of this but we should remove the `HINTS` below from the `find_package` calls too.

ekilmer (changes requested):

pgoodman (approved):

CMakeLists.txt
tetsuo-cpp: [19c4be8](https://github.com/lifting-bits/anvill/pull/339/commits/19c4be86a4a252bc1f556fc2982ed93f67a7306a)

Dockerfile
```diff
@@ -30,7 +30,6 @@ ARG ARCH
 ARG LLVM_VERSION
 ARG CXX_COMMON_VERSION
 ARG LIBRARIES
-ARG REMILL_COMMIT_ID=master
```
tetsuo-cpp, line 33: This argument is harder to support with submodules and it didn't seem very important to me so I just removed it. If someone wants to build with a different commit, they can just modify the submodule.

.remill_commit_id
tetsuo-cpp: Good point. How's this?

CMakeLists.txt
```diff
@@ -22,14 +22,12 @@ endif(ANVILL_ENABLE_INSTALL)
...
 
-set(_VCPKG_SHARE_DIR "${Z_VCPKG_ROOT_DIR}/installed/${VCPKG_TARGET_TRIPLET}/${CMAKE_INSTALL_DATADIR}")
-
-find_package(XED CONFIG REQUIRED HINTS "${_VCPKG_SHARE_DIR}")
-find_package(gflags CONFIG REQUIRED HINTS "${_VCPKG_SHARE_DIR}")
-find_package(glog CONFIG REQUIRED HINTS "${_VCPKG_SHARE_DIR}")
+find_package(XED CONFIG REQUIRED)
+find_package(gflags CONFIG REQUIRED)
+find_package(glog CONFIG REQUIRED)
 find_package(Z3 4.8 CONFIG REQUIRED)
```
pgoodman, line 28: I think latest Z3 in vcpkg is past 4.8, and so you might be able to remove this version now too.

pgoodman (approved):

CMakeLists.txt
tetsuo-cpp: [ee1d7a1](https://github.com/lifting-bits/anvill/pull/339/commits/ee1d7a18d41ab12a36492b47836f04de93081932)

Yep, VCPKG has 4.11 so this can go.

ekilmer: As an aside, version handling in CMake is a bit complicated. It _could_ be used to specify that 4.8 is the minimum compatible version. Still, since I don't think we've tested that and we're not aiming for compatibility outside of vcpkg (at least for now), then removal of this is probably best.

See here for all the details of CMake `find_package` version searching https://cmake.org/cmake/help/latest/command/find_package.html#config-mode-version-selection

ekilmer (approved):
LGTM! Love seeing more deletions than additions 😄

pgoodman (approved):

PR #340, opened 2022-12-21
PowerPC Support

tetsuo-cpp:


include/anvill/Lifters.h
```diff
@@ -248,8 +248,7 @@ class LifterOptions {
   //      (add (ptrtoint __anvill_pc) <address>)
   //
   // Otherwise, a concrete integer is used, i.e. `<address>`.
-  std::function<llvm::Value *(llvm::IRBuilderBase &, const remill::Register *,
-                              uint64_t)>
+  std::function<llvm::Value *(llvm::IRBuilderBase &, llvm::Type *, uint64_t)>
```
lkorenc, line 251: Type alias.

lib/Arch/PPC_C.cpp
```diff
@@ -0,0 +1,332 @@
...
+  err = BindParameters(func, injected_sret, fdecl.params);
+  if (remill::IsError(err)) {
+    return err;
+  }
+
+  fdecl.return_stack_pointer_offset = 0;
+  fdecl.return_stack_pointer = arch->RegisterByName("R1");
+
+  fdecl.return_address.reg = arch->RegisterByName("LR");
+  fdecl.return_address.type = fdecl.return_address.reg->type;
```
lkorenc, line 134: Wrap in a method? Sets fixed data except for one thing (which can do as arg).

lib/Arch/PPC_C.cpp
```diff
@@ -0,0 +1,332 @@
...
+  return llvm::Error::success();
+}
+
+llvm::Error
+PPC_C::BindReturnValues(llvm::Function &function, bool &injected_sret,
+                        std::vector<anvill::ValueDecl> &ret_values) {
+
+  llvm::Type *ret_type = function.getReturnType();
+  LOG(INFO) << "Binding on return " << remill::LLVMThingToString(ret_type);
+  injected_sret = false;
```
lkorenc, line 145: Modifies argument even if result is failure, is this desired? Consider using tuple as return type (since bool is easily default constructible).

lib/Arch/PPC_C.cpp
```diff
@@ -0,0 +1,332 @@
...
+  fdecl.return_address.type = fdecl.return_address.reg->type;
+
+  return llvm::Error::success();
+}
+
+llvm::Error
+PPC_C::BindReturnValues(llvm::Function &function, bool &injected_sret,
+                        std::vector<anvill::ValueDecl> &ret_values) {
+
+  llvm::Type *ret_type = function.getReturnType();
```
lkorenc, line 143:
```suggestion
  auto ret_type = function.getReturnType();
```

lib/Arch/PPC_C.cpp
```diff
@@ -0,0 +1,332 @@
...
+          "Function '%s' with sret-attributed parameter has non-void return type '%s'",
+          function.getName().str().c_str(),
+          remill::LLVMThingToString(ret_type).c_str());
+    }
+
+    // NOTE(alex): Not sure about this, check later.
+    //
+    // Indirect return values are passed by pointer through `R3`.
+    value_declaration.reg = arch->RegisterByName("R3");
+    return llvm::Error::success();
```
lkorenc, line 165: Shouldn't this set `injected_sret` to `true`?

lib/Arch/PPC_C.cpp
```diff
@@ -0,0 +1,332 @@
...
+    // FPRs
+    RegisterConstraint({VariantConstraint("F1", kTypeFloat, kMaxBit32)}),
+    RegisterConstraint({VariantConstraint("F2", kTypeFloat, kMaxBit32)}),
+    RegisterConstraint({VariantConstraint("F3", kTypeFloat, kMaxBit32)}),
+    RegisterConstraint({VariantConstraint("F4", kTypeFloat, kMaxBit32)}),
+    RegisterConstraint({VariantConstraint("F5", kTypeFloat, kMaxBit32)}),
+    RegisterConstraint({VariantConstraint("F6", kTypeFloat, kMaxBit32)}),
+    RegisterConstraint({VariantConstraint("F7", kTypeFloat, kMaxBit32)}),
+    RegisterConstraint({VariantConstraint("F8", kTypeFloat, kMaxBit32)}),
+    RegisterConstraint({VariantConstraint("F9", kTypeFloat, kMaxBit32)}),
```
Ninja3047, line 47: https://math-atlas.sourceforge.net/devel/assembly/elfspec_ppc.pdf
3-14 says parameter passing only goes up to F8

lib/Arch/PPC_C.cpp
```diff
@@ -0,0 +1,332 @@
...
+    RegisterConstraint({VariantConstraint("F13", kTypeFloat, kMaxBit32)}),
+};
+
+static const std::vector<RegisterConstraint> kReturnRegConstraints = {
+    // GPRs
+    RegisterConstraint({VariantConstraint("R3", kTypeIntegral, kMaxBit32)}),
+    RegisterConstraint({VariantConstraint("R4", kTypeIntegral, kMaxBit32)}),
+    // FPRs
+    RegisterConstraint({VariantConstraint("F1", kTypeFloat, kMaxBit32)}),
+    RegisterConstraint({VariantConstraint("F2", kTypeFloat, kMaxBit32)}),
```
Ninja3047, line 60: F1 is only used for return value

lib/Arch/PPC_C.cpp
```diff
@@ -0,0 +1,332 @@
...
+    RegisterConstraint({VariantConstraint("R3", kTypeIntegral, kMaxBit32)}),
+    RegisterConstraint({VariantConstraint("R4", kTypeIntegral, kMaxBit32)}),
+    RegisterConstraint({VariantConstraint("R5", kTypeIntegral, kMaxBit32)}),
+    RegisterConstraint({VariantConstraint("R6", kTypeIntegral, kMaxBit32)}),
+    RegisterConstraint({VariantConstraint("R7", kTypeIntegral, kMaxBit32)}),
+    RegisterConstraint({VariantConstraint("R8", kTypeIntegral, kMaxBit32)}),
+    RegisterConstraint({VariantConstraint("R9", kTypeIntegral, kMaxBit32)}),
+    RegisterConstraint({VariantConstraint("R10", kTypeIntegral, kMaxBit32)}),
+    // FPRs
+    RegisterConstraint({VariantConstraint("F1", kTypeFloat, kMaxBit32)}),
```
Ninja3047, line 39: should these be max 32 bit or max 64bit?

lib/Arch/PPC_C.cpp
```diff
@@ -0,0 +1,332 @@
...
+
+#include "AllocationState.h"
+#include "Arch.h"
+
+namespace anvill {
+namespace {
+
+static const std::vector<RegisterConstraint> kParamRegConstraints = {
+    // GPRs
+    RegisterConstraint({VariantConstraint("R3", kTypeIntegral, kMaxBit32)}),
```
Ninja3047, line 30: this seems right but would be nice to provide a reference or something that says the max value is 32 bit even though the registers support 64bit width for our particular processor

lib/Arch/PPC_C.cpp
```diff
@@ -0,0 +1,332 @@
...
+#include <remill/Arch/Arch.h>
+#include <remill/Arch/Name.h>
+
+#include "AllocationState.h"
+#include "Arch.h"
+
+namespace anvill {
+namespace {
+
+static const std::vector<RegisterConstraint> kParamRegConstraints = {
```
Ninja3047, line 28: probably should include a link or two to the docs you used for reference in a comment

Ninja3047 (commented):
looks good so far
probably should add some tests with the PPC challenge binaries

Ninja3047:
tests need to be re-ran to pull in latest irene3 after PPC changes are merged in

Ninja3047:
Also i think our remill branch needs to also be based off master instead of the patchable-ir-branch

tetsuo-cpp:
> probably should add some tests with the PPC challenge binaries

If you have a moment, could you please help me with this? Looks like it needs to be added so some JSON configuration but I haven't looked at it deeply.

lib/Arch/PPC_C.cpp
tetsuo-cpp: Hmm, what are you suggesting to be wrapped in a method?

lib/Arch/PPC_C.cpp
tetsuo-cpp: [5e99152](https://github.com/lifting-bits/anvill/pull/340/commits/5e99152bc1fbc258ae6dcca43e0a96186137640b)

That variable is removed so no longer applies.

lib/Arch/PPC_C.cpp
tetsuo-cpp: [5e99152](https://github.com/lifting-bits/anvill/pull/340/commits/5e99152bc1fbc258ae6dcca43e0a96186137640b)

include/anvill/Lifters.h
tetsuo-cpp: [12bbc6d](https://github.com/lifting-bits/anvill/pull/340/commits/12bbc6dc49b2630c2252a065d824e4a45f1686dc)

lib/Arch/PPC_C.cpp
tetsuo-cpp: That doc you linked seems to suggest that it's also used for passing parameters.

lib/Arch/PPC_C.cpp
tetsuo-cpp: Yeah good pick up. The GPRs should have 64 too.

lib/Arch/PPC_C.cpp
tetsuo-cpp: [88a3367](https://github.com/lifting-bits/anvill/pull/340/commits/88a33677d5f7ce6b9a6020c661ff9c600f5cbe23)

lib/Arch/PPC_C.cpp
tetsuo-cpp: [88a3367](https://github.com/lifting-bits/anvill/pull/340/commits/88a33677d5f7ce6b9a6020c661ff9c600f5cbe23)

lib/Arch/PPC_C.cpp
tetsuo-cpp: [88a3367](https://github.com/lifting-bits/anvill/pull/340/commits/88a33677d5f7ce6b9a6020c661ff9c600f5cbe23)

lib/Arch/PPC_C.cpp
tetsuo-cpp: [88a3367](https://github.com/lifting-bits/anvill/pull/340/commits/88a33677d5f7ce6b9a6020c661ff9c600f5cbe23)

I actually can't find anything that talks about this specifically but I think it's sort of self-explanatory as our toolchain conforms to the 32-bit EABI which is standard on a lot of PPC devices with 32-bit GPRs. I've just linked to the toolchain documentation, not sure if that's good enough.

lib/Arch/PPC_C.cpp
Ninja3047: ah yeah, misworded
i meant to say that F2-F4 aren't used as return registers
looks like you fixed this

tetsuo-cpp: Oh right, understood.

tetsuo-cpp:
@2over12 Since we've bumped the `common_cxx` version, we'll need to remove `build_linux (ubuntu, 20.04, headless, 3.1.3479, 15, v0.2.16` from the list of required checks. I don't have admin privileges over Anvill, are you able to do this?

Ninja3047:
hrmm, we're failing some tests still even though it should be running on the old test set
will take a look

tetsuo-cpp:
There's still some non-Go tests failing. I'll take a look at those today.

2over12:
> 

I'll look at having a task setup with a stable name as a required check

Ninja3047 (approved):
lgtm

tetsuo-cpp:
@Ninja3047 Thanks for fixing the build. [38f6461](https://github.com/lifting-bits/anvill/pull/340/commits/38f6461504624d1ab5f5ee9fb19c4a2dc6b80217) suggests that there's still a race here that needs fixing. Could we document that in an issue?

ci/challenge_bins_test_settings.json
```diff
@@ -1,21 +1,37 @@
...
+        "challenge-3_x86_program_go",
         "challenge-3_armv7_program_go.elf",
-        "challenge-3_amd64_program_go_patched.elf",
-        "challenge-3_amd64_program_go.elf",
-        "challenge-3_x86_program_go_patched.elf",
-        "challenge-3_x86_program_go.elf"
+        "challenge-3_armv7_program_go_patched.elf",
+        "challenge-10-ppc-program_c.elf",
+        "challenge-10-ppc-vle-program_c.elf",
+        "ppc-adc_mpc5744p.elf",
```
2over12, line 16: Should we be ignoring ppc here? I'd like to have at least one ppc-binary running in CI to merge this

2over12 (changes requested):

ci/challenge_bins_test_settings.json
Ninja3047: pushed a couple ppc hello binaries to the old test set
these were passing locally so they hopefully pass CI
challenge 10 ppc doesn't work atm

2over12 (approved):

PR #343, opened 2023-01-12
Fix CI Build

2over12:


.github/workflows/build.yml
```diff
@@ -143,8 +143,8 @@ jobs:
       - name: Install stable rust
```
Ninja3047, line 143:
```
- uses: dtolnay/rust-toolchain@stable
```
might be better here for installing the rust toolchain + setting up path

Ninja3047 (changes requested):

Ninja3047 (approved):
lgtm (assuming passes CI)

Ninja3047 (approved):
lgtm

PR #345, opened 2023-02-02
Update CI

Ninja3047:
This fixes tests
Also drop llvm-14 + binja + update cxx-common


tetsuo-cpp:
This needs to have `LLVM_VERSION` checks removed. Pushing that up now.

Ninja3047:
This will likely need a bit of work since i expect a bunch of the new binaries to fail

.github/workflows/build.yml
```diff
@@ -52,12 +52,11 @@ jobs:
...
-        binja:
-          - { channel: "headless", version: "3.1.3479" }
-        llvm: ["14", "15"]
-        cxxcommon_version: ["v0.2.16"]
+        llvm: ["15"]
+        cxxcommon_version: ["v0.2.22"]
 
-    runs-on: ubuntu-20.04
+    runs-on:
+      labels: gha-ubuntu-32
```
ekilmer, line 59: A larger runner is necessary?

2over12 (approved):
LGTM, can you PR your changes to lifting-tools-ci when you have a chance to go along with this PR [libraries/lifting-tools-ci](https://github.com/lifting-bits/anvill/pull/345/files#diff-9008814c34e1c32b3e98cbeadf6cde1c005f4f77a0dc20c1be424ef772fd8d9b)

.github/workflows/build.yml
Ninja3047: we have like almost 200 test binaries that need to run through (and a bunch more that are disabled) so i figured using a larger one would be better for this case

scripts/test-amp-challenge-bins.sh
```diff
@@ -109,7 +109,7 @@ do
 done
 
 FAILED="no"
-for dir in challenge-binaries
```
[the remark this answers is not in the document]
Ninja3047: the script used to generate the binaries (https://github.com/trailofbits/IRENE-challenges/blob/main/build-challenges.sh#L3) uses a different directory for some reason

this has the added benefit of allowing us test using both sets so it doesn't break other WIP PRs or the master branch

2over12 (approved):
Let's get this merged

PR #347, opened 2023-02-08
Stop unnecessarily parsing the semantic module for each basic block

tetsuo-cpp:


lib/Lifters/BasicBlockLifter.cpp
```diff
@@ -42,7 +42,6 @@ CallableBasicBlockFunction BasicBlockLifter::LiftBasicBlockFunction() && {
 
 
   this->RecursivelyInlineFunctionCallees(bbfunc.func);
-  anvill::EntityLifter lifter(options);
```
tetsuo-cpp, line 45: This constructor loads the semantic module for the given architecture and is responsible for a good chunk of the runtime. I'm assuming since `lifter` doesn't get used, we can just remove this.

2over12 (approved):
Good find, lgtm

PR #348, opened 2023-02-09
Move DSE to the end of optimization for performance

tetsuo-cpp:


tetsuo-cpp:
Hmm, the plugin timeouts don't seem to be happening anymore. Not sure what that's about.

Ninja3047:
> Hmm, the plugin timeouts don't seem to be happening anymore. Not sure what that's about. Perhaps there's something in #345 that's causing it.

yeah in #345 we updated the set of tests for CI which isn't in the `patchable-ir-main` branch
this will need to be based on `will/update-build` so it gets the update set of tests

2over12:
Do we have stats/benchmarks on what this is looking like performance-wise? I could see arguments going both ways in terms of perf and I also want to check that we arent regressing code quality by accidentally blocking someting

tetsuo-cpp:
> Do we have stats/benchmarks on what this is looking like performance-wise? I could see arguments going both ways in terms of perf and I also want to check that we arent regressing code quality by accidentally blocking someting

I have a comment explaining how this change affects the performance of Anvill and the brevity of the output: https://github.com/trailofbits/irene3/issues/117#issuecomment-1423720498. Although I just did `wc -l` on the IR file so I realize that's pretty superficial. Were you thinking of something more detailed than that?

tetsuo-cpp:
Hmm, CI doesn't seem to be running.

tetsuo-cpp:
As mentioned in https://github.com/trailofbits/irene3/issues/117#issuecomment-1428672702, the `anvill-decompile` invocations don't seem to be timing out anymore. I think this is good to go.

frabert (approved):

2over12 (approved):
Alright I've validated the code quality on a couple of challenge problem examples

2over12:
> @2over12 @frabert @Ninja3047 As mentioned in [trailofbits/irene3#117 (comment)](https://github.com/trailofbits/irene3/issues/117#issuecomment-1428672702), the `anvill-decompile` invocations don't seem to be timing out anymore. I think this is good to go.

Is this because I disabled timeout failures to get https://github.com/lifting-bits/anvill/pull/343, I would be surprised if these changes resolved everything

tetsuo-cpp:
> Is this because I disabled timeout failures to get #343 merged with a plan to re-enable them, I would be surprised if these changes resolved everything

I've been grepping for `Anvill Decompile returned -131` lines in the output so I don't think that the `lifting-tools-ci` change should affect that.

tetsuo-cpp:
It's kind of hard to tell which binary is causing the OOM. However, just running a few challenges locally, and looking at memory usage with `top` and Instruments, I'm not seeing any worse memory usage with these performance fixes.

From my observations, the pattern tends to be that memory usage goes up steadily during the lifting phase since we're adding more and more nodes, instructions, etc to our LLVM module. Then, it flattens out a bit when we start optimizations.

My theory is that there's probably nothing in this changeset that is increasing memory usage to any significant degree. Before, Anvill was probably not getting far enough into some of these workloads to hit the memory limit before getting killed due to timeout.

I'm thinking that I might make a change to `lifting-tools-ci` to set a memory limit for each spawned subprocess. That way, each subprocess that requests too much memory will fail to allocate and fail (and we can debug each one) rather than letting GitHub Actions manager thingy step in and send a SIGKILL to the entire script.

2over12:
> It's kind of hard to tell which binary is causing the OOM. However, just running a few challenges locally, and looking at memory usage with `top` and Instruments, I'm not seeing any worse memory usage with these performance fixes.
> 
> From my observations, the pattern tends to be that memory usage goes up steadily during the lifting phase since we're adding more and more nodes, instructions, etc to our LLVM module. Then, it flattens out a bit when we start optimizations.
> 
> My theory is that there's probably nothing in this changeset that is increasing memory usage to any significant degree. Before, Anvill was probably not getting far enough into some of these workloads to hit the memory limit before getting killed due to timeout.
> 
> @2over12 I'm thinking that I might make a change to `lifting-tools-ci` to set a memory limit for each spawned subprocess. That way, each subprocess that requests too much memory will fail to allocate and fail (and we can ignore failures and debug each one) rather than letting GitHub Actions manager thingy step in and send a SIGKILL to the entire script. Does that seem ok to you?

Sounds reasonable to me, you'll probably need to switch how decompilation tasks are spawned to multiprocessing in order to enforce process isolation which may lead to same issues with IPC. I don't want to keep fighting with CI here given time constraints so if this approach becomes unreasonable we can also consider skipping the problematic challenges for now

tetsuo-cpp:
CI is failing but it looks the same as what's happening in `patchable-ir-main`. Refer to https://github.com/trailofbits/irene3/issues/117#issuecomment-1436129084 about why the reduced verification of modules change didn't make its way into this patch.

I'll merge this now.

tetsuo-cpp:
@2over12 Looks like we might not really need this anymore since I'm about to give CI an overhaul. I'm thinking that I'll bring across the `DLOG` fixes since those are valid regardless, but we can leave pass ordering as is.

2over12 (approved):
LGTM

PR #351, opened 2023-02-14
Add global register support

Ninja3047:


include/anvill/Declarations.h
```diff
@@ -117,11 +117,18 @@ struct OffsetDomain {
   ValueDecl target_value;
   std::int64_t stack_offset;
 };
+
+struct ConstantDomain {
+  ValueDecl target_value;
+  std::uint64_t value;
+};
+
+typedef std::variant<OffsetDomain, ConstantDomain> ValueDomain;
```
2over12, line 126: I dont see this type def used anywhere

lib/Protobuf.cpp
```diff
@@ -624,17 +623,32 @@ void ProtobufTranslator::ParseCFGIntoFunction(
...
+      } else {
+        LOG(FATAL) << symval.curr_val().GetTypeName()
+                   << " is unimplemented for affine relations";
+      }
     }
 
-    SpecStackOffsets off = {affine_equalities};
+    SpecStackOffsets off = {stack_offsets};
     decl.stack_offsets.insert({blk_addr, off});
+    decl.constant_values.insert({blk_addr, constant_values});
```
2over12, line 651: Maybe moves and emplaces here in retrospect?

2over12 (changes requested):
LGTM minor nits

lib/Lifters/BasicBlockLifter.cpp
```diff
@@ -481,12 +481,23 @@ BasicBlockFunction BasicBlockLifter::CreateBasicBlockFunction() {
...
   };
 
-
   LOG(INFO) << "Live values at entry to function "
             << this->block_context->LiveBBParamsAtEntry().size();
   this->UnpackLiveValues(ir, ptr_provider, this->state_ptr,
                          this->block_context->LiveBBParamsAtEntry());
 
+  for (auto &reg_const : this->block_context->GetConstants()) {
+    auto new_value = this->options.program_counter_init_procedure(
```
2over12, line 490: If it's not too much effort we probably want to make this conditioned on a flag here: https://github.com/trailofbits/irene3/blob/4381e2cf8d318628d88a9c80a0fe02310b09268f/data_specifications/specification.proto#L160 ie. the spec should specify wether to construct a plain integer constant or not. If this isn't too much effort for you would be good to have, otherwise we can merge this and Ill address it in my upcoming rework of equalities

2over12 (approved):
One thing, can either address here or address later

lib/Lifters/BasicBlockLifter.cpp
2over12: Ive decided that we'll address this in the tail call representation PR

2over12:
Good to merge once CI passes

2over12 (approved):
LGTM

PR #353, opened 2023-02-16
Update globals ci

Ninja3047:
update lifting-tools-ci to run global register fixup script
hopefully CI works first try


Ninja3047:
Done in #355

PR #354, opened 2023-02-17
Adds a control flow simplification pass to support idiomatic block gotos

2over12:
This is a temporary measure and will be superseded by #349. For now we call an intrinsic that signals a goto to address

### DIFF
--- a/lib/Passes/RemoveAssignmentsToNextPC.cpp
+++ b/lib/Passes/RemoveAssignmentsToNextPC.cpp
@@ -1,0 +1,109 @@
+#include <anvill/Passes/RemoveAssignmentsToNextPC.h>
+#include <llvm/IR/Constants.h>
+#include <llvm/IR/DerivedTypes.h>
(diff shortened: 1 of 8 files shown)


lib/Passes/RemoveAssignmentsToNextPC.cpp
```diff
@@ -0,0 +1,135 @@
...
+#include <llvm/IR/PassManager.h>
+#include <llvm/IR/PatternMatch.h>
+#include <llvm/IR/Verifier.h>
+#include <llvm/Support/Casting.h>
+#include <llvm/Support/raw_ostream.h>
+#include <remill/BC/Util.h>
+
+#include <optional>
+
+#include "anvill/ABI.h"
```
frabert, line 19: `<anvill/ABI.h>`

frabert (changes requested):
LGTM other than the header include for consistency

frabert (approved):

PR #356, opened 2023-02-23
Rework load lifted value and store native to operate over composite low level storage

2over12:
Expands the generality of all valuedecls to allow splitting across multiple registers etc


2over12:
Currently disables the ABI building with a stub to avoid updating all of them since we don't declare functions using the ABI since they are specified

frabert (approved):

PR #357, opened 2023-02-24
Force pack lives to occur before recovered anvill returns when replacing remill returns

2over12:
This is a hack to better support the semantics of returns, in reality what we want here is for every block exiting control flow effect to be preceded by a live variable pack, maybe this can be achieved in #349

### DIFF
--- a/lib/Passes/RemoveAssignmentsToNextPC.cpp
+++ b/lib/Passes/RemoveAssignmentsToNextPC.cpp
@@ -17,6 +17,8 @@
 
 #include <optional>
 
+#include "Utils.h"
+
 namespace anvill {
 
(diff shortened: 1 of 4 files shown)


lib/Passes/RemoveAssignmentsToNextPC.cpp
```diff
@@ -17,6 +17,8 @@
 
 #include <optional>
 
+#include "Utils.h"
```
Ninja3047, line 20: extra include?

Ninja3047 (approved):
small nit

lib/Passes/RemoveAssignmentsToNextPC.cpp
2over12: it's for the private pass utils

PR #358, opened 2023-02-24
Recover externals' names

frabert:
Keeping values used until the optimizations run help recover the external references.

Some calls are still getting optimized away (especially when the return values are not used), but I haven't been able to figure out why that happens


2over12:
I think this is superseded by #356 ?

frabert:
🤦 yeah forgot to test after merging that one

PR #365, opened 2023-03-16
Adds back splitting the stack around the return address preventing exposing the return address pointer 

2over12:


frabert (approved):
LGTM

PR #366, opened 2023-03-17
Metrics for Reference Recovery

Ninja3047:
* Always enable these statistics
 * Collect loads/stores to __anvill_sp

### DIFF
--- a/lib/Passes/CodeQualityStatCollector.cpp
+++ b/lib/Passes/CodeQualityStatCollector.cpp
@@ -59,20 +59,17 @@ class ConditionalComplexityVisitor
 
 
 llvm::PreservedAnalyses
-CodeQualityStatCollector::run(llvm::Function &function,
-                              llvm::FunctionAnalysisManager &analysisManager) {
+CodeQualityStatCollector::run(llvm::Module &module,
+                              llvm::ModuleAnalysisManager &analysisManager) {
   ConditionalComplexityVisitor complexity_visitor;
-  llvm::GlobalVariable* anvill_sp = function.getParent()->getGlobalVariable(kSymbolicSPName);
-  llvm::GlobalVariable* anvill_pc = function.getParent()->getGlobalVariable(kSymbolicPCName);
+  llvm::GlobalVariable* anvill_sp = module.getGlobalVariable(kSymbolicSPName);
+  llvm::GlobalVariable* anvill_pc = module.getGlobalVariable(kSymbolicPCName);
 
   if (anvill_sp != nullptr) {
     for (const auto &U: anvill_sp->uses()) {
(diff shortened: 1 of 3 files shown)


lib/Passes/CodeQualityStatCollector.cpp
```diff
@@ -62,15 +63,26 @@ CodeQualityStatCollector::run(llvm::Function &function,
...
     }
 
+    if (auto *store_inst = llvm::dyn_cast<llvm::StoreInst>(&i)) {
+      if (store_inst->getPointerOperand()->getName() == kSymbolicSPName) {
+        ++AnvillStackPointers;
+      }
+    }
+
+    if (auto *load_inst = llvm::dyn_cast<llvm::LoadInst>(&i)) {
+      if (load_inst->getPointerOperand()->getName() == kSymbolicSPName) {
```
2over12, line 80: This is too specific. A load inst may derive a pointer from `inttoptr( add (ptrtoint  anvill_sp) 15)`. You need to chase down an expression that built up a pointer to see if it references anvill_sp. There is infrastructure to do that https://github.com/lifting-bits/anvill/blob/dd89a2316c1d5ade3640aba5f727dc93c2605b67/include/anvill/Utils.h#L79 . We perhaps could also do a similar thing with this pass before anvill_pc removal to see if we missed pc relative references. Regardless of if we use that approach or some other approach we need a way to check if we missed globals as well

lib/Passes/CodeQualityStatCollector.cpp
```diff
@@ -11,13 +11,14 @@
 #include <llvm/IR/PassManager.h>
 
 namespace anvill {
-STATISTIC(
+ALWAYS_ENABLED_STATISTIC(
```
2over12, line 14: Why do we want this?

2over12 (changes requested):

lib/Passes/CodeQualityStatCollector.cpp
```diff
@@ -57,32 +59,61 @@ class ConditionalComplexityVisitor
...
+  if (anvill_sp != nullptr) {
+    for (const auto &U: anvill_sp->uses()) {
+      const auto &user = U.getUser();
+      if (const llvm::Instruction *I = llvm::dyn_cast<llvm::Instruction>(user)) {
+        sp_funcs.insert(I->getFunction());
+      }
     }
+  }
 
+  AnvillStackPointers = sp_funcs.size();
```
2over12, line 80: As a pass, this probably isn't safe since this can be run on multiple modules, should probably be an add. Not the end of the world though

2over12 (approved):
Small nit related to running on multiple modules

PR #367, opened 2023-03-29
Ian/fix hash for typespecs

2over12:
Fixes up hashes so they match `operator==()`


frabert:
LLVM already provides `llvm::hash_combine` which I think should be used in the future

PR #369, opened 2023-04-04
Debug utils

Ninja3047:
add script to help with debugging llvm opt passes


bin/Decompile/Main.cpp
```diff
@@ -89,6 +92,24 @@ int main(int argc, char *argv[]) {
   google::ParseCommandLineFlags(&argc, &argv, true);
   google::InitGoogleLogging(argv[0]);
 
+  if (FLAGS_llvm_debug) {
+    llvm::DebugFlag = true;
+  }
+
+  if (FLAGS_llvm_print_before_all || FLAGS_llvm_print_after_all) {
+    std::vector<const char*> fake_argv;
```
2over12, line 100: Not to split hairs on debug code but could we not assign verbose to https://github.com/llvm/llvm-project/blob/8dfdcc7b7bf66834a761bd8de445840ef68e4d1a/llvm/include/llvm/IR/PrintPasses.h#L30 and just have it print changed to avoid this fake argparsing stuff, or as an alternative I  guess we could pass llvm options through

2over12 (approved):
no big deal if you decide not to do the mentioned change

tetsuo-cpp (approved):
Nice, LGTM.

2over12 (approved):

PR #376, opened 2023-04-27
Get Anvill building with LLVM 16

tetsuo-cpp:


lib/Optimize.cpp
```diff
@@ -204,7 +204,7 @@ void OptimizeModule(const EntityLifter &lifter, llvm::Module &module,
   fpm.addPass(llvm::VerifierPass());
   fpm.addPass(llvm::DSEPass());
   fpm.addPass(llvm::VerifierPass());
-  fpm.addPass(llvm::SROAPass());
+  fpm.addPass(llvm::SROAPass(llvm::SROAOptions::ModifyCFG));
```
tetsuo-cpp, line 207: This pass now takes in some configuration (the other option is called `PreserveCFG` and restricts what optimizations are allowed).

lib/Passes/SliceManager.cpp
```diff
@@ -57,14 +57,13 @@ void SliceManager::insertClonedSliceIntoFunction(
   auto bb = llvm::BasicBlock::Create(this->mod.get()->getContext(),
                                      "slicebasicblock." + std::to_string(id.id),
                                      targetFunc);
+  llvm::IRBuilder<> builder(bb);
 
-  std::for_each(slice.begin(), slice.end(), [bb](llvm::Instruction *insn) {
-    bb->getInstList().push_back(insn);
-  });
+  std::for_each(slice.begin(), slice.end(),
+                [&builder](llvm::Instruction *insn) { builder.Insert(insn); });
```
tetsuo-cpp, line 63: Sorry for the noise here. This file hasn't been Clang formatted for a while I guess.

This is the main change in this file. We don't have raw access to the instruction list anymore so I need to do it like this.

We can have an `#if` that calls the new API when we're building with LLVM 16, but I believe the `IRBuilder` is pretty lightweight to construct so I think this is fine.

tetsuo-cpp:
There's some Irene build problem that's causing CI to fail. But other than that, it looks ok.

pgoodman:
If/when that unrelated bug is fixed, can you migrate to cxx-common 0.3.2 (unless a newer one comes out in the meantime), and drop llvm 15 support?

tetsuo-cpp:
> If/when that unrelated bug is fixed, can you migrate to cxx-common 0.3.2 (unless a newer one comes out in the meantime), and drop llvm 15 support?

Do you want to drop LLVM 15 support across the board? In my Remill change, I've still kept it backwards compatible with 15. I think being compatible with <15 is hard because of opaque pointers, but I don't think 15+ is much of a burden.

tetsuo-cpp:
Again, this CI failure is happening in `patchable-ir-main` so we can merge this when we're happy with it.

pgoodman:
> Do you want to drop LLVM 15 support across the board?

I think the hold-out might be Rellic after this? Otherwise, when LLVM 17 is out, I'd want to definitely drop 15. Ideally, I'd like to stay as-up-to-date as possible, and drop old dependencies as soon as we can (i.e. as soon as all of our stuff is migrated).

tetsuo-cpp:
> > Do you want to drop LLVM 15 support across the board?
> 
> I think the hold-out might be Rellic after this? Otherwise, when LLVM 17 is out, I'd want to definitely drop 15. Ideally, I'd like to stay as-up-to-date as possible, and drop old dependencies as soon as we can (i.e. as soon as all of our stuff is migrated).

Ok let's drop 15 support here. Remill will still need to support 15 since it also needs to be compatible with Anvill `master`.

tetsuo-cpp:
@pgoodman @2over12 Friendly ping, can we look at getting this in?

pgoodman (approved):

PR #379, opened 2023-05-18
Add image base spec

Ninja3047:


include/anvill/Specification.h
```diff
@@ -124,6 +124,12 @@ class Specification {
   // Return the architecture used by this specification.
   std::shared_ptr<const remill::Arch> Arch(void) const;
 
+  // Return the image name used by this specification.
+  std::string ImageName(void) const;
```
tetsuo-cpp, line 128: Should return a const ref here to avoid copying on every call.

lib/Specification.cpp
```diff
@@ -32,8 +32,10 @@ namespace anvill {
 
 SpecificationImpl::~SpecificationImpl(void) {}
 
-SpecificationImpl::SpecificationImpl(std::unique_ptr<const remill::Arch> arch_)
+SpecificationImpl::SpecificationImpl(std::unique_ptr<const remill::Arch> arch_, std::string image_name_, std::uint64_t image_base_)
```
tetsuo-cpp, line 35: Should either take a const ref or do `std::move` below. Otherwise two copies happen here.

lib/Specification.cpp
```diff
@@ -326,8 +338,11 @@ Specification::DecodeFromPB(llvm::LLVMContext &context, const std::string &pb) {
     return arch.Error();
   }
 
+  auto image_name = spec.image_name();
```
tetsuo-cpp, line 341: One more copy here which you can avoid with `const auto&`.

tetsuo-cpp (approved):
LGTM! Just a few nits. I'll push up a commit and merge this.

Edit: Or not. Looks like someone other than me has to approve again since I pushed last.

lib/Specification.cpp
tetsuo-cpp: [ccbac6a](https://github.com/lifting-bits/anvill/pull/379/commits/ccbac6a4749f767dabadc244e8739256f5c5e46a)

lib/Specification.cpp
tetsuo-cpp: [ccbac6a](https://github.com/lifting-bits/anvill/pull/379/commits/ccbac6a4749f767dabadc244e8739256f5c5e46a)

include/anvill/Specification.h
tetsuo-cpp: [ccbac6a](https://github.com/lifting-bits/anvill/pull/379/commits/ccbac6a4749f767dabadc244e8739256f5c5e46a)

data_specifications/specification.proto
```diff
@@ -378,11 +378,13 @@ message ControlFlowOverrides {
...
   Arch arch = 1;
   OS operating_system = 2;
-  repeated Function functions = 3;
-  repeated Callsite callsites = 4;
-  repeated GlobalVariable global_variables = 5;
-  repeated Symbol symbols = 6;
-  repeated MemoryRange memory_ranges = 7;
-  ControlFlowOverrides overrides = 8;
-  map<int64, TypeSpec> type_aliases = 9;
+  string image_name = 3;
```
2over12, line 381: just as a general thing for compatibility please in the future do not shift fields when adding to the protobuf def. add a new field with a new index

2over12 (approved):
LGTM

PR #386, opened 2023-07-07
Add `required_globals` to spec

frabert:


lib/Specification.h
```diff
@@ -92,6 +92,8 @@ class SpecificationImpl
   std::vector<Misc> misc_overrides;
 
   std::unordered_map<std::uint64_t, ControlFlowOverride> control_flow_overrides;
+
+  std::vector<std::string> required_globals;
```
2over12: is there any reason for this not to be a set?

2over12 (approved):

lib/Specification.h
frabert: Fixed!

PR #388, opened 2023-08-04
Use encoded type for constant value mappings

tetsuo-cpp:
This was what was causing `DECOMPILE_MODE` to remain in the decompilation. Basically, the expected type differed and because of that, our code didn't know to ignore the "live" value.

These constants aren't necessarily stack pointers so we should use the type information in the spec instead of making that assumption.


Ninja3047 (approved):
lgtm

PR #389, opened 2023-08-07
Remove challenge testing

tetsuo-cpp:
This has been superseded by the Irene3 nightly metrics run so let's remove this to speed up our PR testing.


2over12 (changes requested):
Hmm maybe we want to preserve some basic smoketests just to have something to catch really broken PRs? Maybe a really small anghabench sample?

2over12:
going to merge this but I'd like to add back a small number of angha bench targets per arch to just catch small bugs

2over12 (approved):

PR #390, opened 2023-09-18
Ian/allow for typehints in spec

2over12:
Adds typehints on values at a given address. The basic strategy here is to store an intrinsic hint at any low value after that instruction. The pointer lifter then converts intrinsics into metadata it can use.

### DIFF
--- a/lib/Lifters/BasicBlockLifter.cpp
+++ b/lib/Lifters/BasicBlockLifter.cpp
@@ -11,7 +11,6 @@
 #include <llvm/IR/LLVMContext.h>
 #include <llvm/IR/Value.h>
 #include <llvm/IR/Verifier.h>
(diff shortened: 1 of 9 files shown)


lib/Lifters/BasicBlockLifter.cpp
```diff
@@ -8,9 +8,11 @@
 #include <llvm/IR/DerivedTypes.h>
 #include <llvm/IR/Function.h>
 #include <llvm/IR/IRBuilder.h>
+#include <llvm/IR/Instruction.h>
 #include <llvm/IR/LLVMContext.h>
```
frabert: Check that these includes are actually necessary

frabert (approved):
LGTM, maybe just double check the added includes to be actually necessary

PR #396, opened 2023-09-29
UID codeblocks refactor

ekilmer:


lib/Lifters/BasicBlockLifter.cpp
```diff
@@ -402,13 +402,16 @@ void BasicBlockLifter::LiftInstructionsIntoLiftedFunction() {
 }
 
 
-llvm::MDNode *BasicBlockLifter::GetBasicBlockAnnotation(uint64_t addr) const {
+llvm::MDNode *BasicBlockLifter::GetBasicBlockAddrAnnotation(uint64_t addr) const {
```
2over12: is there any reason to keep this around?

lib/Lifters/FunctionLifter.cpp
```diff
@@ -335,22 +336,25 @@ llvm::Function *FunctionLifter::DeclareFunction(const FunctionDecl &decl) {
   return GetOrDeclareFunction(decl);
 }
 
-BasicBlockLifter &FunctionLifter::GetOrCreateBasicBlockLifter(uint64_t addr) {
-  std::pair<uint64_t, uint64_t> key{curr_decl->address, addr};
+static Uid GetRandUid() {
```
2over12: Do we use this anywhere?

lib/Lifters/FunctionLifter.cpp
```diff
@@ -497,7 +501,11 @@ llvm::Function *FunctionLifter::LiftFunction(const FunctionDecl &decl) {
   //
   // TODO: This could be a thunk, that we are maybe lifting on purpose.
   //       How should control flow redirection behave in this case?
-  auto &entry_lifter = this->GetOrCreateBasicBlockLifter(this->func_address);
+  auto &cfg = this->curr_decl->cfg;
+  auto blk = std::find_if(cfg.begin(), cfg.end(),
```
2over12: Maybe for safety we should have the spec declare the entry block uid to avoid situations where we create a block and an empty block as the first block in a function

2over12 (changes requested):

lib/Lifters/FunctionLifter.cpp
ekilmer: It is not used. Will remove.

lib/Lifters/BasicBlockLifter.cpp
ekilmer: Ah! I forgot to remove this. I removed all call-sites but forgot to delete the implementation.

lib/Lifters/FunctionLifter.cpp
ekilmer: Good idea. Will change this now.

2over12 (approved):
lgtm

PR #398, opened 2023-10-03
Callsite override

Ninja3047:


lib/Passes/RemoveCallIntrinsics.cpp
```diff
@@ -43,23 +38,45 @@ RemoveCallIntrinsics::runOnIntrinsic(llvm::CallInst *remillFunctionCall,
       ra.references_global_value ||  // Related to a global var/func.
       ra.references_program_counter) {  // Related to `__anvill_pc`.
 
-    // TODO(Ian): ignoring callsite decls for now
-    auto fdecl = spec.FunctionAt(ra.u.address);
+    std::optional<CallableDecl> maybe_decl;
```
2over12: Shouldnt this be initialized to spec.FunctionAt folded address ie. the default is we just call the address and dont have to rely on the override if one doesnt exist?

2over12 (changes requested):

lib/Passes/RemoveCallIntrinsics.cpp
Ninja3047: https://github.com/lifting-bits/anvill/pull/398/commits/4e83da7f9ef3e614bf1b41af54ed9a6b262d1b18

2over12 (approved):

lib/Specification.cpp
```diff
@@ -394,6 +395,16 @@ Specification::DecodeFromPB(llvm::LLVMContext &context, std::istream &pb) {
   return Specification(std::move(pimpl));
 }
 
+// Return the call site at a given function address, instruction address pair, or an empty `shared_ptr`.
+std::shared_ptr<const CallSiteDecl>
+Specification::CallSiteAt(std::pair<std::uint64_t, std::uint64_t> loc) const {
```
tetsuo-cpp: Don't need to copy here: `const std::pair<std::uint64_t, std::uint64_t>&`

lib/Passes/RemoveCallIntrinsics.cpp
```diff
@@ -43,23 +38,42 @@ RemoveCallIntrinsics::runOnIntrinsic(llvm::CallInst *remillFunctionCall,
...
+        auto block_contexts = spec.GetBlockContexts();
+        if (auto bb_ctx_ref =
+                block_contexts.GetBasicBlockContextForAddr(*bb_addr);
+            bb_ctx_ref.has_value()) {
+          const auto &bb_ctx = bb_ctx_ref->get();
+          auto func = bb_ctx.GetParentFunctionAddress();
+          if (auto override_decl = spec.CallSiteAt({func, *pc_val})) {
+            DLOG(INFO) << "Overriding call site at " << std::hex << *pc_val
+                       << " in " << std::hex << func;
+            callable_decl = override_decl;
```
tetsuo-cpp: Shared pointer doesn't need to be copied: `callable_decl = std::move(override_decl)`

lib/Passes/RemoveCallIntrinsics.cpp
```diff
@@ -43,23 +38,42 @@ RemoveCallIntrinsics::runOnIntrinsic(llvm::CallInst *remillFunctionCall,
...
+        spec.FunctionAt(ra.u.address);
+
+    if (auto pc_val =
+            GetMetadata(lifter.Options().pc_metadata_name, *remillFunctionCall);
+        pc_val.has_value()) {
+      if (auto bb_addr = GetBasicBlockAddr(f); bb_addr.has_value()) {
+        auto block_contexts = spec.GetBlockContexts();
+        if (auto bb_ctx_ref =
+                block_contexts.GetBasicBlockContextForAddr(*bb_addr);
+            bb_ctx_ref.has_value()) {
```
tetsuo-cpp: I'm not super familiar with these helpers, but is it expected that we can get a None here given that we know we're looking at a basic block function?

lib/Utils.cpp
```diff
@@ -244,6 +244,34 @@ std::string CreateVariableName(std::uint64_t addr) {
   return ss.str();
 }
 
+std::optional<std::uint64_t> GetMetadata(llvm::StringRef tag,
+                                         llvm::Instruction &instr) {
```
tetsuo-cpp: `const llvm::Instruction &`

tetsuo-cpp (approved):
LGTM codewise modulo some C++ nits, I'm missing context around this change though. What's this solving?

Ninja3047:
> LGTM codewise modulo some C++ nits, I'm missing context around this change though. What's this solving?

the main use case is for fixing varargs functions
right now when we see a varargs function, we don't necessarily correctly detect the number of parameters actually used. ghidra has the ability to override a function call at a particular call site but we weren't using that information before. this pr makes it so we fetch that information (which was already in the spec) and apply it in the remove call intrinsics pass.

tetsuo-cpp (approved):
Awesome, thanks for the added context.

PR #399, opened 2023-10-14
Fix accesses to globals that occur in the middle of the global

2over12:


lib/Passes/RewriteVectorOps.cpp
```diff
@@ -0,0 +1,228 @@
...
+    return std::nullopt;
+  }
+
+
+  std::optional<RewrittenInteger> ConsumeNext() {
+    uint32_t start_index = this->curr_index;
+    int first_end = sv.getMaskValue(this->curr_index);
+    int prev_ind = first_end;
+    this->curr_index += 1;
+    LOG(INFO) << "first: " << first_end;
```
frabert: `LOG(DEBUG)` or `LOG(TRACE)`

lib/Passes/RewriteVectorOps.cpp
```diff
@@ -0,0 +1,228 @@
...
+    while (!this->ConsumedAll()) {
```
frabert, line 91: I would maybe add a comment here explaining what this loop is accomplishing

frabert (changes requested):

frabert (approved):
LGTM!

PR #400, opened 2023-10-17
Add pass removing shufflevectors

2over12:
actually tested this with alive... dont really want to add alive as a dependency


2over12:
we can probably let https://github.com/lifting-bits/anvill/pull/399 supersede this

2over12:
Superseded by #399

PR #403, opened 2023-11-14
Fix infinite loop with self-referential global struct

ekilmer:
The fix allows the following program to be lifted.

```c
// gcc -O0 -g -gdwarf-4 self_refer.c -o self_refer

#include <stdio.h>

struct foo {
  int num;
  struct foo	*foo_next;
};

static struct foo foo0 = {
  .num = 3,
  .foo_next = &foo0
};

int main() {
  printf("Num: %d\n", foo0.foo_next->num);
}
```


Ninja3047 (approved):
seems like a reasonable fix to me

2over12 (approved):
lgtm

PR #408, opened 2024-02-28
Invalid func fix

Ninja3047:
Return nullptr if the cfg does not contain the entry block. This is similar to what we return if the address is not valid or the memory is not executable.


2over12 (approved):
lgtm

PR #409, opened 2024-02-29
Invalid func fix pt2

Ninja3047:
Return the decl instead of a nullptr

### DIFF
--- a/lib/Lifters/FunctionLifter.cpp
+++ b/lib/Lifters/FunctionLifter.cpp
@@ -409,13 +409,6 @@ llvm::Function *FunctionLifter::LiftFunction(const FunctionDecl &decl) {
     return nullptr;
   }
 
-  // Function has no valid instructions.
-  auto &cfg = decl.cfg;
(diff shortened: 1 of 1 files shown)


2over12 (approved):

lib/Lifters/FunctionLifter.cpp
```diff
@@ -440,6 +433,14 @@ llvm::Function *FunctionLifter::LiftFunction(const FunctionDecl &decl) {
     return native_func;
   }
 
+  // Function has no valid instructions.
+  auto &cfg = decl.cfg;
+  if (cfg.find(decl.entry_uid) == cfg.end()) {
+    LOG(WARNING) << "Function missing entry block " << std::hex << decl.address;
+    return native_func;
```
2over12: i think we should still return null (ie. fail) but only if it's not a decl

if the client wants to declare incoherent functions that should be on the client

2over12 (changes requested):

2over12 (approved):

